PR #286, opened 2023-08-15
Disable token authentication for API v3 publishes

richardhuaaa:
The next version of the protocol will no longer use identity keys, and instead uses installation keys. Installation keys are also of a different format (Ed25519 + Curve25519) vs identity keys (secp256k1). We can temporarily disable authentication for ~V1~ V3 publishes during development.

~Regarding API versioning - what are our thoughts on naming this API version 3, instead of 1, to align with the client versioning?~

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -35,7 +35,7 @@ type Config struct {
 // Options bundle command line options associated with the authn package.
 type AuthnOptions struct {
 	Enable              bool     `long:"enable" description:"require client authentication via wallet tokens"`
-	EnableV3            bool     `long:"enable-v3" description:"require client authentication via wallet tokens for V3"`
+	EnableV3            bool     `long:"enable-v3" description:"require client authentication for V3"`
 	Ratelimits          bool     `long:"ratelimits" description:"apply rate limits per wallet"`
 	AllowLists          bool     `long:"allowlists" description:"apply higher limits for allow listed wallets (requires authz and ratelimits)"`
 	PrivilegedAddresses []string `long:"privileged-address" description:"allow this address to publish into other user's topics"`


pkg/api/interceptor.go
```diff
@@ -67,9 +67,23 @@ func (wa *WalletAuthorizer) Stream() grpc.StreamServerInterceptor {
...
 }
 
+func (wa *WalletAuthorizer) isProtocolVersion3(request *messagev1.PublishRequest) bool {
+	envelopes := request.Envelopes
+	if envelopes == nil || len(envelopes) == 0 {
+		return false
+	}
+	// If any of the envelopes are not for a v1 topic, then we treat the request as non-v1
+	for _, envelope := range envelopes {
+		if !strings.HasPrefix(envelope.ContentTopic, "/xmtp/3/") && !strings.HasPrefix(envelope.ContentTopic, "xmtp/3/") {
```
snormore, line 77: The `xmtp/3` prefix variant should never really exist, or if messages come in that way then I think shouldn't be treated as a valid `/xmtp/3` protocol message. For example, if messages came in on `xmtp/0` the cleaner would wipe them after a day because it's not `/xmtp/0`

pkg/api/authentication_test.go
```diff
@@ -19,6 +19,61 @@ func Test_AuthnNoToken(t *testing.T) {
 	})
 }
 
+func Test_AuthnNoTokenNonV1(t *testing.T) {
```
snormore, line 22: The test names still mention v1 too, should be v3

pkg/api/config.go
```diff
@@ -35,6 +35,7 @@ type Config struct {
 // Options bundle command line options associated with the authn package.
 type AuthnOptions struct {
 	Enable              bool     `long:"enable" description:"require client authentication via wallet tokens"`
+	EnableV3            bool     `long:"enable-v3" description:"require client authentication via wallet tokens for V3"`
```
neekolas, line 38: The description seems wrong. This doesn't require any wallet tokens

pkg/api/interceptor.go
```diff
@@ -67,9 +67,23 @@ func (wa *WalletAuthorizer) Stream() grpc.StreamServerInterceptor {
...
 }
 
+func (wa *WalletAuthorizer) isProtocolVersion3(request *messagev1.PublishRequest) bool {
+	envelopes := request.Envelopes
+	if envelopes == nil || len(envelopes) == 0 {
+		return false
+	}
+	// If any of the envelopes are not for a v3 topic, then we treat the request as non-v3
+	for _, envelope := range envelopes {
+		if !strings.HasPrefix(envelope.ContentTopic, "/xmtp/3/") {
```
neekolas, line 77: Why if any of the envelopes are for V3. This would allow someone to get around our auth system by including a random V3 message with every request. 

I think we should change this to a requirement that _all_ envelopes must be V3

richardhuaaa: If a request is mixed V0 and V3, and the enableV3 flag is not on, then that would disable authentication for V0 requests, which is not what we want, right? See `Test_AuthnNoTokenMixedV0V3`

PR #288, opened 2023-08-23
Add Subscribe2 method

neekolas:
## Summary

- Updates to latest version of `xmtp/proto`
- Adds new GRPC-only method `Subscribe2` which supports bidirectional streaming
- Updates testing infrastructure to support `Subscribe2`

## Testing
- I have run the `xmtp-js` test suite against an image running this and nothing breaks, and I have tested the new API manually with `bot-kit-pro`


pkg/api/message/v1/dispatcher.go
```diff
@@ -79,6 +79,43 @@ func (d *dispatcher) Unregister(ch chan interface{}, topics ...string) {
 	d.unregister(ch, topics...)
 }
 
+// Update takes in an array of topics and unregisters any topics that are not in the array and registers any topics that are not already registered.
+func (d *dispatcher) Update(ch chan interface{}, topics ...string) {
```
neekolas, line 83: I don't strictly need to create this method. I _could_ just unregister everything and then re-register everything.

I added this method for two reasons:
1. Messages could get lost if a message got published during the unregister process and was waiting for the lock
2. For large numbers of subs (we have people subscribing to 10k+ topics at a time), it's going to be a lot more expensive to add and remove all those listeners.

I could be convinced that I'm over-thinking this and we should just simplify.

pkg/api/message/v1/service.go
```diff
@@ -167,6 +168,58 @@ func (s *Service) Subscribe(req *proto.SubscribeRequest, stream proto.MessageApi
...
+	for {
+		select {
+		case <-stream.Context().Done():
+			log.Debug("stream closed")
+			return nil
+		case <-s.ctx.Done():
+			log.Info("service closed")
+			return nil
+		case req := <-requestChannel:
+			log.Info("updating subscription", zap.Int("num_content_topics", len(req.ContentTopics)))
```
neekolas, line 206: I'm not logging the full list of topics in this new method. We should think about removing that in the main subscription method too. The subscription list is sometimes thousands of items long, and has little value in the logs.

mkobetic: Would it make sense to log the topics added/removed instead?

pkg/api/message/v1/dispatcher.go
mkobetic: I think we definitely do want the smart update behavior, the only thing I'm wondering if it's worth having add/remove subscriptions request, given that the lists could be 10k+ long.

pkg/api/message/v1/dispatcher.go
```diff
@@ -79,6 +79,54 @@ func (d *dispatcher) Unregister(ch chan interface{}, topics ...string) {
...
+		return
+	}
+
+	if len(toUnregister) > 0 {
+		d.Unregister(ch, toUnregister...)
+	}
+
+	toRegister := make([]string, 0)
+
+	// Lock again
```
mkobetic, line 117: It looks like `Register()` handles already registered topics. You only need to figure out what to unregister here.

mkobetic (approved):
Few nitpicks but overall LGTM

pkg/api/message/v1/service.go
neekolas: Still would be a very long list for the initial call of a popular user/bot

pkg/api/message/v1/dispatcher.go
neekolas: It's a reasonable question. It's nice in that there is less data going over the wire. But it does put more work onto the client to maintain a list of subscriptions and figure out the delta. Because there is no ackowledgement that an update has succeeded, this way also gives us a "self-healing" property where if some update isn't successfully processed the next update will set things right.

PR #289, opened 2023-08-23
IP based rate limiting

mkobetic:
I think this is the smallest change to achieve what we agreed on in https://github.com/xmtp-labs/hq/issues/1101. Here is the summary of the rules with the new changes in place (I can summarize the old rules, but I'm not sure that's really relevant at this point). None of this applies unless `--api.authn.enable` is set:

## `--api.authn.ratelimits`

All requests are subject to rate limits. Requests are bucketed by client IP address (there is one bucket for all requests without IPs). Each bucket is allocated a number of tokens that are refilled at a fixed rate per minute up to a given maximum number of tokens. Requests cost 1 token by default, except Publish requests cost the number of Envelopes carried and BatchQuery requests cost the number of queries carried. If `--api.authn.allowlists` is also set then requests with Bearer tokens from wallets explicitly `Allowed` get priority, i.e. higher bucket token max and refill rate. Priority wallets get separate IP buckets from regular wallets.

## `--api.authn.allowlists`

All requests that require authentication (currently Publish only) will be rejected for wallets that are explicitly `Denied`. Normally they will be permitted according to the rules of the request type, i.e. you can't publish into other wallets' contact and private topics (this applies regardless of this option).

Wallets that are explicitly `Allowed` will get priority rate limits if `--api.authn.ratelimits` is set.

## Notes

* The authorization flow is rewired a fair bit, I tried to disentangle the authorization and rate limiting bits, hopefully it's a bit clearer now.
* The rate limiter changes are 
  * adding a notion of cost to the spend logic
  * reifying Limit handling to make overriding the defaults easier for testing
  * also tried to clean up the terminology, calling things priority instead of allow listed, since that's a bit confusing in the rate limiting context, also purging the term wallet
* Added `Allow, Apply` to the AllowLister to be able to test priority rate limiting
* It may make sense to have completely different limits for stream requests given that they are long lasting and tax the system differently.
* ~~Also to replicate what we have at the edge we should have different limits for publishes and for the other requests.~~

### DIFF
--- a/pkg/api/interceptor.go
+++ b/pkg/api/interceptor.go
@@ -189,7 +189,7 @@ func (wa *WalletAuthorizer) applyLimits(ctx context.Context, fullMethod string, 
 	if err == nil {
 		return nil
 	}
-	wa.Log.Debug("rate limited",
+	wa.Log.Info("rate limited",
 		logging.String("client_ip", ip),
 		logging.WalletAddress(wallet.String()),
 		logging.Bool("priority", isPriority),
--- a/pkg/api/interceptor.go
+++ b/pkg/api/interceptor.go
@@ -13,6 +13,7 @@ import (
 
 	messagev1 "github.com/xmtp/proto/v3/go/message_api/v1"
 	"github.com/xmtp/xmtp-node-go/pkg/logging"
+	"github.com/xmtp/xmtp-node-go/pkg/ratelimiter"
 	"github.com/xmtp/xmtp-node-go/pkg/types"
 )
 
@@ -45,9 +46,22 @@ func (wa *WalletAuthorizer) Unary() grpc.UnaryServerInterceptor {
 		info *grpc.UnaryServerInfo,
 		handler grpc.UnaryHandler,
 	) (interface{}, error) {
+		if !wa.Ratelimits && !wa.requiresAuthorization(req) {
+			return handler(ctx, req)
+		}
+		wallet, authErr := wa.getWallet(ctx)
+
+		if wa.Ratelimits {
+			if err := wa.applyLimits(ctx, info.FullMethod, req, wallet); err != nil {
+				return nil, err
+			}
+		}
 
 		if wa.requiresAuthorization(req) {
-			if err := wa.authorize(ctx, req); err != nil {
+			if authErr != nil {
+				return nil, status.Error(codes.Unauthenticated, authErr.Error())
+			}
+			if err := wa.authorize(ctx, req, wallet); err != nil {
 				return nil, err
 			}
 		}
@@ -62,14 +76,20 @@ func (wa *WalletAuthorizer) Stream() grpc.StreamServerInterceptor {
 		info *grpc.StreamServerInfo,
 		handler grpc.StreamHandler,
 	) error {
-		// TODO(mk): Add metrics
+		if wa.Ratelimits {
+			ctx := stream.Context()
+			wallet, _ := wa.getWallet(ctx)
+			if err := wa.applyLimits(ctx, info.FullMethod, nil, wallet); err != nil {
+				return err
+			}
+		}
 		return handler(srv, stream)
 	}
 }
 
 func (wa *WalletAuthorizer) isProtocolVersion3(request *messagev1.PublishRequest) bool {
 	envelopes := request.Envelopes
-	if envelopes == nil || len(envelopes) == 0 {
+	if len(envelopes) == 0 {
 		return false
 	}
 	// If any of the envelopes are not for a v3 topic, then we treat the request as non-v3
@@ -86,34 +106,37 @@ func (wa *WalletAuthorizer) requiresAuthorization(req interface{}) bool {
 	return isPublish && (!wa.isProtocolVersion3(publishRequest) || wa.AuthnConfig.EnableV3)
 }
 
-func (wa *WalletAuthorizer) authorize(ctx context.Context, req interface{}) error {
+func (wa *WalletAuthorizer) getWallet(ctx context.Context) (types.WalletAddr, error) {
 	md, ok := metadata.FromIncomingContext(ctx)
 	if !ok {
-		return status.Errorf(codes.Unauthenticated, "metadata is not provided")
+		return "", status.Errorf(codes.Unauthenticated, "metadata is not provided")
 	}
 
 	values := md.Get(authorizationMetadataKey)
 	if len(values) == 0 {
-		return status.Errorf(codes.Unauthenticated, "authorization token is not provided")
+		return "", status.Errorf(codes.Unauthenticated, "authorization token is not provided")
 	}
 
 	words := strings.SplitN(values[0], " ", 2)
 	if len(words) != 2 {
-		return status.Errorf(codes.Unauthenticated, "invalid authorization header")
+		return "", status.Errorf(codes.Unauthenticated, "invalid authorization header")
 	}
 	if scheme := strings.TrimSpace(words[0]); scheme != "Bearer" {
-		return status.Errorf(codes.Unauthenticated, "unrecognized authorization scheme %s", scheme)
+		return "", status.Errorf(codes.Unauthenticated, "unrecognized authorization scheme %s", scheme)
 	}
 	token, err := decodeAuthToken(strings.TrimSpace(words[1]))
 	if err != nil {
-		return status.Errorf(codes.Unauthenticated, "extracting token: %s", err)
+		return "", status.Errorf(codes.Unauthenticated, "extracting token: %s", err)
 	}
 
 	wallet, err := validateToken(ctx, wa.Log, token, time.Now())
 	if err != nil {
-		return status.Errorf(codes.Unauthenticated, "validating token: %s", err)
+		return "", status.Errorf(codes.Unauthenticated, "validating token: %s", err)
 	}
+	return wallet, nil
+}
 
+func (wa *WalletAuthorizer) authorize(ctx context.Context, req interface{}, wallet types.WalletAddr) error {
 	if pub, isPublish := req.(*messagev1.PublishRequest); isPublish {
 		for _, env := range pub.Envelopes {
 			if !wa.privilegedAddresses[wallet] && !allowedToPublish(env.ContentTopic, wallet) {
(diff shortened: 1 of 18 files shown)


pkg/api/interceptor.go
```diff
@@ -62,14 +75,20 @@ func (wa *WalletAuthorizer) Stream() grpc.StreamServerInterceptor {
 		info *grpc.StreamServerInfo,
 		handler grpc.StreamHandler,
 	) error {
-		// TODO(mk): Add metrics
+		if wa.Ratelimits {
```
mkobetic, line 78: It may make sense to have completely different limits for stream requests given that they are long lasting and tax the system differently.

mkobetic:
I have yet to write some tests, but I think this is ready for feedback at this point.

mkobetic:
ok, basic tests are in now.

pkg/ratelimiter/rate_limiter.go
```diff
@@ -31,88 +31,98 @@ type Entry struct {
...
 
+// Limit controls token refilling for bucket entries
+type Limit struct {
+	// Maximum number of tokens that can be accumulated
+	MaxTokens uint16
+	// Number of tokens to refill per minute
+	RatePerMinute uint16
+}
+
+func (l Limit) Refill(entry *Entry) {
```
mkobetic, line 42: This was just extracted from fillAndReturnEntry

pkg/api/interceptor.go
```diff
@@ -86,62 +105,86 @@ func (wa *WalletAuthorizer) requiresAuthorization(req interface{}) bool {
...
+	var priority bool
+	if len(wallet) > 0 && wa.AllowLists {
+		priority = wa.AllowLister.IsAllowListed(wallet.String())
+	}
+
+	// default request cost is 1,
+	// note that nil req (subscription stream) has cost 1 as well
+	cost := 1
+	switch req := req.(type) {
+	case *messagev1.PublishRequest:
```
neekolas, line 175: It looks like this will run against `Query` requests as well as `Publish`. If that's the case, we probably want to make the cost much higher for publish. For example, the WAF rules offer 10X the capacity for querying. Maybe we increase the rate of tokens and make publish cost 10 per envelope?

pkg/ratelimiter/rate_limiter.go
```diff
@@ -10,15 +10,15 @@ import (
...
 
 const (
-	ALLOW_LISTED_RATE_PER_MINUTE = uint16(10)
-	ALLOW_LISTED_MAX_TOKENS      = uint16(10000)
-	REGULAR_RATE_PER_MINUTE      = uint16(1)
-	REGULAR_MAX_TOKENS           = uint16(100)
-	MAX_UINT_16                  = 65535
+	PRIORITY_RATE_PER_MINUTE = uint16(10)
+	PRIORITY_MAX_TOKENS      = uint16(10000)
+	REGULAR_RATE_PER_MINUTE  = uint16(1)
```
neekolas, line 15: We should adjust these rates to match the current WAF limits as best we can. Current WAF limits are 200 publishes + 2k queries/minute.

So, if we adjusted the cost of publishing to 10 per publish it would be 4000/minute with a `REGULAR_MAX_TOKENS` of 20k. 

That feels high, since a spammer could just use their entire allocation on publishing and none on querying and get 2000/5 minutes/node. Makes me wonder if we actually want to track the limits separately depending on request type.

pkg/ratelimiter/rate_limiter.go
```diff
@@ -31,88 +31,98 @@ type Entry struct {
...
+	mutex         sync.RWMutex
+	PriorityLimit Limit
+	RegularLimit  Limit
 }
 
 func NewTokenBucketRateLimiter(log *zap.Logger) *TokenBucketRateLimiter {
 	tb := new(TokenBucketRateLimiter)
 	tb.log = log.Named("ratelimiter")
-	tb.wallets = make(map[string]*Entry)
+	tb.buckets = make(map[string]*Entry)
```
neekolas, line 73: It occurs to me that we never clean up this map. We should at least make a todo to periodically purge expired entries, otherwise it'll slowly eat up all the system memory.

pkg/api/interceptor.go
```diff
@@ -45,9 +45,22 @@ func (wa *WalletAuthorizer) Unary() grpc.UnaryServerInterceptor {
 		info *grpc.UnaryServerInfo,
 		handler grpc.UnaryHandler,
 	) (interface{}, error) {
+		if !wa.Ratelimits && !wa.requiresAuthorization(req) {
```
neekolas, line 48: @richardhuaaa recently added an escape hatch for V3 messages, because they don't have auth tokens. But now that we are rate limiting by IP, we can include them again.

pkg/ratelimiter/rate_limiter.go
mkobetic: Agreed, I raised the same questions in the Notes. I also wonder if stream requests should be limited differently given their distinct nature. I can definitely look into separating the limits, I agree to address different requests types with single limit will have weird edge cases.

pkg/api/interceptor.go
mkobetic: Yup, we can still keep the flag to avoid authorizing V3 publishes and then we can let them through the same limit flow. They won't be able to get priority limits, but that may be fine in the short term.

pkg/ratelimiter/rate_limiter.go
mkobetic: If we do different limits for different request types, do we want different priority limits too?

neekolas: I'd suggest a blanket 5X increase for all rate limits if you have priority status. That feels high enough to be worth doing, but not so high they are going to risk network stability.

My vote is for separate limits per RPC. Especially now that this repo is open source, people will figure out how to game the combined limits very quickly.

pkg/api/interceptor.go
neekolas: Considering there are maybe 5 addresses in the entirety of the V3 address space right now, I think we're a long way from needing to worry about priority limits.

pkg/ratelimiter/rate_limiter.go
mkobetic: OK, so let me spell it out in details. Per request limits ( trying to match the edge limits ) :

| RPC Type | rate / max |
| ---------- | --------------|
| Query, BatchQuery(per query) | 2000/10000 |
| Publish (per envelope) | 200 / 1000 |
| Subscribe/Subscribe2/SubscribeAll | 2000 / 10000 |

Priority limits will be fixed multiple of the regular limits. We can make the multiplier configurable and default to 5.

I'd be tempted to dial the streaming calls down a fair bit, but DD shows that we are averaging about twice as many subscribe call (in prod) than query calls over the past week.

Screenshot 2023-08-25 at 14 01 41

mkobetic: I should note that limit `2000/10000` allows a spike of 18000 in the first 5 minutes assuming full token budget, but it can't be sustained over longer time, since you have 0 tokens at the end of it and have to stick with the 2000 refill rate per minute.

mkobetic:
We also need rate limiting metrics for observability. Either in this PR or as a follow up.

pkg/ratelimiter/rate_limiter.go
neekolas: I think that sounds reasonable as a starting place. I'd rather start with these limits being a little generous, and then dial back, rather than breaking any applications on release.

neekolas: I'm OK with leaving the subscribe calls at the same limits as Query. While you're right that the worst case Subscribe call is much more expensive than a Query, we'd have to gather some data to really know what is more expensive. A Query call always hits the database, and a Subscribe call never does. I'd bet most subscriptions don't receive a single message in their lifetime, so it's really just the overhead of handling the connection and setting up the broadcaster. But a subscription to a popular topic would be quite expensive.

neekolas:
@mkobetic I'm concerned about relying on the `x-forwarded-for` header here. Looking at [this DataDog](https://app.datadoghq.com/logs?query=api%20request%20%40client_ip%3A172.16.62.1&cols=host%2Cservice&index=%2A&messageDisplay=inline&stream_sort=desc&viz=stream&from_ts=1692386956941&to_ts=1692991756941&live=true) page, when we are logging out the API requests 99.9% of them are coming from a single IP in the last week. That makes me think the header is being set by the GRPC Gateway, or some other intermediate service.

mkobetic:
> @mkobetic I'm concerned about relying on the `x-forwarded-for` header here. Looking at [this DataDog](https://app.datadoghq.com/logs?query=api%20request%20%40client_ip%3A172.16.62.1&cols=host%2Cservice&index=%2A&messageDisplay=inline&stream_sort=desc&viz=stream&from_ts=1692386956941&to_ts=1692991756941&live=true) page, when we are logging out the API requests 99.9% of them are coming from a single IP in the last week. That makes me think the header is being set by the GRPC Gateway, or some other intermediate service, in the base case.
> 
> It also appears to be a private IP. I think we're going to need to do some infrastructure work first to get this reliably from the client -> node.

Hm, I'll definitely double check, but we are getting some valid IPs back, see https://app.datadoghq.com/logs?query=api%20request%20&agg_m=count&agg_q=%40client_ip&agg_t=count&index=&sort_m=&sort_t=&top_n=10&top_o=top&viz=toplist&x_missing=true&from_ts=1692388398105&to_ts=1692993198105&live=true

It seems that particular IP is from the e2e tests, which makes sense.

neekolas:
Maybe it is all from E2E. It just seems fishy to me that the highest number of requests we have from any other address over the past week is 12. A single load of xmtp.chat can be hundreds of requests.

Maybe I'm missing something and that logger is only called on a subset of requests, which would explain why there aren't thousands of IPs with thousands of requests each.

mkobetic:
> Maybe it is all from E2E. It just seems fishy to me that the highest number of requests we have from any other address over the past week is 12. A single load of xmtp.chat can be hundreds of requests.
> 
> Maybe I'm missing something and that logger is only called on a subset of requests, which would explain why there aren't thousands of IPs with thousands of requests each.

Yup, definitely something odd going on here. But I do remember there being much more substantial log of these. We used this to debug some issues and identifying traffic coming from the same source.

snormore:
Yep we switched that request logger to only log at debug level unless it's an error (if it's the same one I'm assuming, haven't completely read the comment backlog here yet) 😅

pkg/ratelimiter/rate_limiter.go
```diff
@@ -31,88 +37,106 @@ type Entry struct {
...
 // TokenBucketRateLimiter implements the RateLimiter interface
 type TokenBucketRateLimiter struct {
-	log     *zap.Logger
-	wallets map[string]*Entry
-	mutex   sync.RWMutex
+	log                *zap.Logger
+	buckets            map[string]*Entry
+	mutex              sync.RWMutex
+	PriorityMultiplier uint16
+	Limits             map[LimitType]*Limit
```
mkobetic, line 75: OK, so the Limiter now has the priority multiplier and a set of limit types. There is the DEFAULT limit and now also a PUBLISH limit. This is set up so that we can add more limit types in the future if we decide to dial down the stream request limits. The `Spend` call now has to pass the limit type to apply. I updated the limit values as we discussed (see above). WDYT?

neekolas:
> We switched that api request logging to just log debug level unless it's an errored request (in https://github.com/xmtp/xmtp-node-go/pull/206)

Ohhhhhhhhh

pkg/api/interceptor.go
```diff
@@ -152,7 +153,7 @@ func (wa *WalletAuthorizer) authorize(ctx context.Context, req interface{}, wall
 	return nil
 }
 
-func (wa *WalletAuthorizer) applyLimits(ctx context.Context, req interface{}, wallet types.WalletAddr) error {
+func (wa *WalletAuthorizer) applyLimits(ctx context.Context, fullMethod string, req interface{}, wallet types.WalletAddr) error {
```
mkobetic, line 156: I ended up not using the `fullMethod` yet, but if we decide to introduce limits for different types of stream requests, we will need it because we don't get the unmarshaled request in that case.

pkg/api/interceptor.go
```diff
@@ -163,28 +164,27 @@ func (wa *WalletAuthorizer) applyLimits(ctx context.Context, req interface{}, wa
...
+		limitType = ratelimiter.PUBLISH
 	case *messagev1.BatchQueryRequest:
 		cost = len(req.Requests)
 	}
 	// need to separate the IP buckets between priority and regular wallets
 	var bucket string
-	if priority {
-		bucket = "P" + ip
+	if isPriority {
+		bucket = "P" + ip + string(limitType)
```
mkobetic, line 183: I hope this isn't too subtle, but this is how we get different buckets for different limit types. I can see an argument that this should happen inside the limiter, but then the simple "give me the bucket name" parameter becomes something else ("bucket prefix"?). So that doesn't feel squeeky clean either.

neekolas (approved):
This looks good to me. We should probably deploy it to dev for a day, with the metrics in there, before it makes its way to production so we can get an estimate of the impact.

mkobetic:
OK, so what's left to do here is
* [ ] limit metrics
* [ ] limiter IP map cleanup job
* [ ] doc updates

mkobetic:
OK, it seems that we should get counts for ratelimited requests already from GRPC metrics, the error code should be `codes.ResourceExhausted` (I also fixed up the error returned for requests that require authorization and the wallet token is in some way unusuable as `codes.Unauhtenticated`).

So I'm only adding 2 metrics to monitor the rate limiter buckets maps (tag `name` = buckets1/buckets2), that are both emitted at sweep end:
1. `xmtp_ratelimiter_buckets` - should be a gauge of size of the bucket maps 
2. `xmtp_ratelimiter_entries_deleted` - should be count of entries deleted in sweeps

mkobetic:
OK, I think I'm done with the changes, I'll try devnet deploy tomorrow.

mkobetic:
With the debug logging turned on I can zoom in on the E2E publish ratelimiting. It confirms that the nodes are blocking it about the rate that the E2E dashboards report. The funny thing is that it doesn't seem to be at all evenly distributed among the nodes (https://app.datadoghq.com/logs?saved-view-id=2087937)

Screenshot 2023-08-30 at 00 17 18

The ratelimiter logs allow peeking into the tokens remaining, which also show an odd looking pattern. No idea why it looks this way yet. (https://app.datadoghq.com/logs?saved-view-id=2087929)

Screenshot 2023-08-30 at 00 17 50

snormore:
I guess to fix the problem of e2e requests being rate limited here we'd just have it always use the same identity for it's auth and add a limit override for that one?

snormore (changes requested):
I guess to fix the problem of e2e runner requests being rate limited here we'd just have it always use the same identity for its auth and add a limit override for it. We should do that before merging, or even before deploying this to dev again.

snormore (approved):

PR #290, opened 2023-08-28
Sweep expired rate limiter buckets

mkobetic:
This adds a janitor process to the rate limiter to prevent unbounded growth of the bucket map. My main concern was locking out request flow while iterating over the potentially large bucket map looking for expired entries.

We can definitely do the search under the read lock, and do the removal separately under a write lock. However that is still potentially blocking requests from new IPs. Moreover as soon as a new IP queues up a write lock request, any following read lock requests will be blocked as well (I'm not 100% sure about this but I don't see how else you can prevent write lock requests being preempted by read locks forever).

To address the write lock blocking I split the bucket maps into two. One will be subject to creating new IP entries, and the other will be subject to the expiration sweep. That way the writes and sweeps should not collide. The cost is having to read 2 maps to check for entry presence (and consequently 2 read locks) on every read access, but I think it's worth it to avoid unpredictable sweep pauses. The maps' roles are swapped after every sweep.

## Notes

* We could try to be more clever about keeping the two maps roughly the same size, but maybe let's collect some metrics first to see whether it's worth it

## Update

Also added another mutex at the Limiter level, to protect the swap of the two buffer maps. Otherwise a goroutine that is in the middle of `fillAndReturnEntry` could end up getting the same map in the second check. Consequently we have two layers of mutexes now so we need to be careful about locking them in a fixed order so that we don't deadlock.

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"strings"
 	"sync"
+	"time"
 
 	middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
@@ -87,9 +88,13 @@ func (s *Server) startGRPC() error {
 	stream = append(stream, telemetryInterceptor.Stream())
 
 	if s.Config.Authn.Enable {
+		limiter := ratelimiter.NewTokenBucketRateLimiter(s.Log)
+		// Expire buckets after 6 hours of inactivity,
+		// sweep for expired buckets every hour.
+		go limiter.Janitor(s.ctx, time.Hour, 6*time.Hour)
 		s.authorizer = NewWalletAuthorizer(&AuthnConfig{
 			AuthnOptions: s.Config.Authn,
-			Limiter:      ratelimiter.NewTokenBucketRateLimiter(s.Log),
+			Limiter:      limiter,
 			AllowLister:  s.Config.AllowLister,
 			Log:          s.Log.Named("authn"),
 		})
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -91,6 +91,8 @@ func (s *Server) startGRPC() error {
 		limiter := ratelimiter.NewTokenBucketRateLimiter(s.Log)
 		// Expire buckets after 6 hours of inactivity,
 		// sweep for expired buckets every hour.
+		// Note: entry expiration should be at least some multiple of
+		// maximum limit max / limit rate minutes.
 		go limiter.Janitor(s.ctx, time.Hour, 6*time.Hour)
 		s.authorizer = NewWalletAuthorizer(&AuthnConfig{
 			AuthnOptions: s.Config.Authn,
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -89,11 +89,11 @@ func (s *Server) startGRPC() error {
 
 	if s.Config.Authn.Enable {
 		limiter := ratelimiter.NewTokenBucketRateLimiter(s.Log)
-		// Expire buckets after 6 hours of inactivity,
-		// sweep for expired buckets every hour.
+		// Expire buckets after 1 hour of inactivity,
+		// sweep for expired buckets every 10 minutes.
 		// Note: entry expiration should be at least some multiple of
-		// maximum limit max / limit rate minutes.
-		go limiter.Janitor(s.ctx, time.Hour, 6*time.Hour)
+		// maximum (limit max / limit rate) minutes.
+		go limiter.Janitor(s.ctx, 10*time.Minute, 1*time.Hour)
 		s.authorizer = NewWalletAuthorizer(&AuthnConfig{
 			AuthnOptions: s.Config.Authn,
 			Limiter:      limiter,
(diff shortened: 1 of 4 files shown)


pkg/ratelimiter/buckets.go
```diff
@@ -0,0 +1,80 @@
...
+	// The locking strategy is adapted from the following blog post: https://misfra.me/optimizing-concurrent-map-access-in-go/
+	b.mutex.RLock()
+	currentVal, exists := b.buckets[bucket]
+	b.mutex.RUnlock()
+	if !exists {
+		if !create {
+			return nil
+		}
+		b.mutex.Lock()
+		currentVal, exists = b.buckets[bucket]
```
mkobetic, line 34: We didn't use to do the second check in the original code, but I think it needs to be there.

pkg/ratelimiter/rate_limiter.go
```diff
@@ -103,25 +106,10 @@ func (rl *TokenBucketRateLimiter) fillAndReturnEntry(limitType LimitType, bucket
 	if isPriority {
 		multiplier = rl.PriorityMultiplier
 	}
-	// The locking strategy is adapted from the following blog post: https://misfra.me/optimizing-concurrent-map-access-in-go/
-	rl.mutex.RLock()
```
mkobetic, line 107: This is mostly just pull into Buckets, with one small fix.

pkg/ratelimiter/rate_limiter.go
```diff
@@ -103,25 +106,10 @@ func (rl *TokenBucketRateLimiter) fillAndReturnEntry(limitType LimitType, bucket
...
-		currentVal = &Entry{
-			tokens:   uint16(limit.MaxTokens * multiplier),
-			lastSeen: time.Now(),
-			mutex:    sync.Mutex{},
-		}
-		rl.buckets[bucket] = currentVal
-		rl.mutex.Unlock()
-
-		return currentVal
+	if entry := rl.buckets2.getAndRefill(bucket, limit, multiplier, false); entry != nil {
```
neekolas, line 109: While this looks pretty buttoned up in the implementation right now, it would be helpful to add some tests asserting that an entry is never in both buckets at the same time. 

That feels like the regression I'd be worried about if I was modifying this code.

pkg/api/server.go
```diff
@@ -87,9 +88,13 @@ func (s *Server) startGRPC() error {
 	stream = append(stream, telemetryInterceptor.Stream())
 
 	if s.Config.Authn.Enable {
+		limiter := ratelimiter.NewTokenBucketRateLimiter(s.Log)
+		// Expire buckets after 6 hours of inactivity,
+		// sweep for expired buckets every hour.
+		go limiter.Janitor(s.ctx, time.Hour, 6*time.Hour)
```
mkobetic, line 94: Not quite sure what the settings should be. Thoughts?

neekolas: Given that we've invested all this effort making it non-blocking, I'd suggest doing it more frequently (maybe every 30 mins) to keep memory usage lower

mkobetic: OK, how about this. Given that a bucket is refilled in 5 minutes given our current settings, how about expiration 1 hour and sweep every 10 mins? Might help keeping the maps smaller as well.

pkg/ratelimiter/rate_limiter_test.go
```diff
@@ -114,3 +97,67 @@ func TestSpendConcurrent(t *testing.T) {
...
+
+	// ip2 should still be out of tokens
+	require.Error(t, rl.Spend(DEFAULT, "ip2", 1, false)) // bucket1 refresh
+	// ip3 should have expired now and we should have 2 tokens again
+	require.NoError(t, rl.Spend(DEFAULT, "ip3", 1, false)) // bucket2 add
+	require.NoError(t, rl.Spend(DEFAULT, "ip3", 1, false)) // bucket2 refresh
+	require.Error(t, rl.Spend(DEFAULT, "ip3", 1, false))   // bucket2 refresh
+}
+
+func TestBucketExpirationIntegrity(t *testing.T) {
```
mkobetic, line 140: @neekolas is this good enough? I dropped the janitor goroutine and instead am invoking the sweepAndSwap directly for better determinism. This is arguably covered by ip2 in the prior test, but I'm adding this to focus on a single specific entry.

mkobetic:
Hm, even though the swap of the two Buffers maps is atomic, I think it may require a lock to synchronize with goroutines that are in the middle of fillAndReturnEntry.

neekolas (approved):

pkg/ratelimiter/rate_limiter.go
```diff
@@ -104,10 +105,14 @@ func (rl *TokenBucketRateLimiter) fillAndReturnEntry(limitType LimitType, bucket
 	if isPriority {
 		multiplier = rl.PriorityMultiplier
 	}
+	rl.mutex.RLock()
 	if entry := rl.oldBuckets.getAndRefill(bucket, limit, multiplier, false); entry != nil {
+		rl.mutex.RUnlock()
```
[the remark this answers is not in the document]
mkobetic: It would, but I was avoiding `defer` on these short paths as we do elsewhere.

Issue #291, opened 2023-08-30
Bug: Pubsub intermittently failing


neekolas:
### Describe the bug

We are seeing intermittent failures of pubsub to gossip messages between nodes. This causes messages to not be delivered to subscriptions.
![CleanShot 2023-08-30 at 12 24 47@2x](https://github.com/xmtp/xmtp-node-go/assets/65710/703965a9-7d9e-40c6-9804-d44cf7bc7c72)

This may be an issue with `libp2p-pubsub`, or `go-waku`'s usage of it

### Expected behavior

Subscriptions reliably deliver messages, irrespective of which node the message was initially published to

### Steps to reproduce the bug

Unclear

PR #292, opened 2023-08-30
Configure libp2p gossipsub WithDirectPeers

snormore:
Attempt to mitigate https://github.com/xmtp/xmtp-node-go/issues/291 with https://github.com/libp2p/go-libp2p-pubsub/blob/master/gossipsub.go#L353-L379

> WithDirectPeers is a gossipsub router option that specifies peers with direct
> peering agreements. These peers are connected outside of the mesh, with all (valid)
> message unconditionally forwarded to them. The router will maintain open connections
> to these peers. Note that the peering agreement should be reciprocal with direct peers
> symmetrically configured at both ends.


neekolas (approved):
Why don't we try this in dev first and make sure it doesn't blow anything up

PR #296, opened 2023-09-01
Poll for new messages instead of subscription

neekolas:
## Summary

Temporary hot fix for subscriptions. Instead of relying on Relay, just poll the database.

Had to create some new store methods to use `receiverTimestamp` (server time) instead of sender timestamp, which is what our existing methods use.

## Known issues
- ~This relies on client timestamps, not server timestamps~
- Streaming messages from ephemeral topics won't work (won't fix)
- ~Test suite continues to fail~


pkg/api/message/v1/service.go
```diff
@@ -79,6 +80,38 @@ func NewService(node *wakunode.WakuNode, logger *zap.Logger) (s *Service, err er
...
+			if err != nil {
+				s.log.Error("error querying store", zap.Error(err))
+				continue
+			}
+			if s.dispatcher != nil {
+				for _, msg := range res {
+					s.dispatcher.Submit(msg.ContentTopic, buildEnvelope(msg))
+				}
+			}
+			lastChecked = newLastChecked
```
richardhuaaa, line 109: Wondering if there is a race condition here if a publish and query happen at the same time that might cause occasional messages to be skipped. Would it make sense to use the receiver timestamp of the latest message fetched here?

richardhuaaa (approved):

pkg/api/message/v1/service.go
neekolas: Ya. I definitely share your concerns about races. Even with the last message's timestamp, you could still get missed messages (clock drift between nodes).

I was thinking about adding some sort of padding to the times and just keeping track of recent message IDs to dedupe.

PR #297, opened 2023-09-02
Use local nats for API subscriptions dispatcher

snormore:
Reverts https://github.com/xmtp/xmtp-node-go/pull/296 and replaces the custom dispatcher implementation with a local nats for API subscriptions.

Fixes https://github.com/xmtp/xmtp-node-go/issues/291


nplasterer (approved):
:shipit:

pkg/api/message/v1/service.go
```diff
@@ -195,61 +176,92 @@ func (s *Service) Subscribe(req *proto.SubscribeRequest, stream proto.MessageApi
...
+			err := pb.Unmarshal(msg.Data, &env)
 			if err != nil {
-				log.Error("sending envelope to subscriber", zap.Error(err))
+				log.Error("parsing envelope from bytes", zap.Error(err))
+				return
 			}
+			if topic == natsWildcardTopic && !isValidSubscribeAllTopic(env.ContentTopic) {
+				return
+			}
+			func() {
```
neekolas, line 196: Is this just being wrapped in an anonymous function so we can use defer in it?

snormore: Yep exactly

pkg/api/message/v1/service.go
```diff
@@ -195,61 +176,92 @@ func (s *Service) Subscribe(req *proto.SubscribeRequest, stream proto.MessageApi
...
-			return nil
-		case obj := <-subC:
-			env, ok := obj.(*proto.Envelope)
-			if !ok {
-				log.Warn("non-envelope received on subscription channel", zap.Any("object", obj))
-				continue
-			}
 
-			err := stream.Send(env)
+	var streamLock sync.Mutex
```
neekolas, line 184: Are we worried that the GRPC server can't handle concurrent `Send` operations?

snormore: Originally added this in xmtpd where we run `-race` tests which would fail/complain about accessing stream in here, so it's just remnants of that but seemed reasonable to keep

neekolas (approved):
Worth a try. Local Nats feels a little overkill for our needs, but at least it's something battle tested and reliable.

This is already deployed to dev? Dev subscriptions look like they are working great right now

snormore:
Yep been on dev for ~45 mins. Definitely agree the local nats feels heavy, but arguably reasonable for the added reliability.

PR #299, opened 2023-09-03
Decouple message store from waku

snormore:
Pull over some message store cleanup from https://github.com/xmtp/xmtp-node-go/pull/295 that decouples it from waku. This should also make it much easier to upgrade the go-waku dependency. Functionally everything should be the same, since we were already writing to the store in the API service on publish, with our custom store, and not relying on the waku store protocol directly. The PR also removes unused waku filter and lightpush code.


dev/run
```diff
@@ -6,11 +6,13 @@ AUTHZ_DB_DSN="postgres://postgres:xmtp@localhost:15432/postgres?sslmode=disable"
 NODE_KEY="8a30dcb604b0b53627a5adc054dbf434b446628d4bd1eccc681d223f0550ce67"
 
 go run cmd/xmtpd/main.go \
-    --message-db-connection-string "${MESSAGE_DB_DSN}" \
-    --message-db-reader-connection-string "${MESSAGE_DB_DSN}" \
-    --authz-db-connection-string "${AUTHZ_DB_DSN}" \
     --nodekey "${NODE_KEY}" \
     --metrics \
     --metrics-period 5s \
+    --store.enable \
```
neekolas, line 12: We'll also need an infra PR to update the command arguments to match these changes.

neekolas (approved):

PR #300, opened 2023-09-03
Upgrade to latest waku-org/go-waku

snormore:
Following up on https://github.com/xmtp/xmtp-node-go/pull/299, this PR upgrades to latest waku-org/go-waku and discontinues use of our out-of-date fork.


go.mod
```diff
@@ -13,188 +13,183 @@ require (
...
 	gopkg.in/yaml.v3 v3.0.1 // indirect
-	lukechampine.com/blake3 v1.1.6 // indirect
+	lukechampine.com/blake3 v1.1.7 // indirect
 	mellium.im/sasl v0.2.1 // indirect
+	nhooyr.io/websocket v1.8.7 // indirect
 )
 
-// To update to a commit CCC run go get -d github.com/xmtp/go-waku@CCC and
-// copy the version identifier from the ouput of that
-replace github.com/status-im/go-waku => github.com/xmtp/go-waku v0.0.0-20220810150132-1237437a419a
```
neekolas, line 197: If we lose our fork, I believe there were some metrics changes that will also get lost. I made some changes to the logger setup, and Martin set up some gauges and metrics in `go-waku` that haven't been merged to the upstream.

I think everything else made it upstream, but not sure.

neekolas: I do think it would be great to retire our fork. Just want to make sure we don't lose things like json logging.

snormore: Looking through the [PRs](https://github.com/xmtp/go-waku/pulls?q=is%3Apr+is%3Aclosed) and it seems like it could be just the [network peers metric](https://github.com/xmtp/go-waku/pull/15/files). I don't see any use of it in our alerts, but do see it on a dashboard. I'm not sure it's used much, it looks pretty consistent with the [connected peers](https://github.com/waku-org/go-waku/blob/master/waku/v2/node/connectedness.go#L69-L85) metric on the dashboard.

Confirmed that json logging still works too, we're passing our logger into it via WithLogger.

neekolas (approved):

PR #304, opened 2023-09-04
Consistent e2e auth identity

snormore:
Support passing in auth keys for e2e runner so that it can have a consistent identity for rate limit overrides in https://github.com/xmtp/xmtp-node-go/pull/289#pullrequestreview-1609891161

### DIFF
--- a/cmd/xmtpd-e2e/main.go
+++ b/cmd/xmtpd-e2e/main.go
@@ -41,17 +41,17 @@ func main() {
 
 	apiURL := envVar("XMTPD_E2E_API_URL", remoteAPIURLByEnv[networkEnv])
 
-	walletKeyHex := envVar("XMTPD_E2E_AUTH_WALLET_KEY", "")
-	var walletKey *ecdsa.PrivateKey
-	if walletKeyHex != "" {
-		walletKey, err = crypto.HexToECDSA(walletKeyHex)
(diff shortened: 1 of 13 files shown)


cmd/xmtpd-e2e/main.go
```diff
@@ -41,6 +41,20 @@ func main() {
...
 	apiURL := envVar("XMTPD_E2E_API_URL", remoteAPIURLByEnv[networkEnv])
 
+	walletKeyHex := envVar("XMTPD_E2E_AUTH_WALLET_KEY", "")
+	var walletKey *ecdsa.PrivateKey
+	if walletKeyHex != "" {
+		walletKey, err = crypto.HexToECDSA(walletKeyHex)
+		requireNoError(err)
+	}
+
+	installationKeyHex := envVar("XMTPD_E2E_AUTH_INSTALLATION_KEY", "")
```
neekolas, line 51: Installation is really a V3 term. This would be the identity key for v1/v2

snormore: Will we be able to reuse these wallet/installation keys for the v3 auth identity too (as in will we need to pass in separate keys here for that or reuse these)?

neekolas: No possibility for re-use of the identity keys/installation keys. Totally different curves and signature schemes. Wallets obviously remain the same.

neekolas (approved):

PR #320, opened 2023-10-23
Validation service scaffold

neekolas:
## Summary

- Adds support for the Rust-backed Validation Service over GRPC
- Adds database models for Installations and KeyPackages
- Adds migrations for those database models
- Adds Service to read/write KeyPackages
-


pkg/api/message/v3/service.go
```diff
@@ -45,11 +48,50 @@ func (s *Service) Close() {
...
+	ids := req.InstallationIds
+	keyPackages, err := s.mlsStore.ConsumeKeyPackages(ctx, ids)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to consume key packages: %s", err)
+	}
+
+	resPackages := make([]*proto.ConsumeKeyPackagesResponse_KeyPackage, len(keyPackages))
+	for _, keyPackage := range keyPackages {
+		// Return the key packages in the original order
+		targetIndex := indexOf(keyPackage.InstallationId, ids)
```
snormore, line 82: Want to build a map of id to index above this loop and do the O(1) lookup here instead of this O(n) indexOf?

pkg/api/message/v3/service.go
```diff
@@ -61,7 +103,27 @@ func (s *Service) PublishWelcomes(ctx context.Context, req *proto.PublishWelcome
...
+	validationResults, err := s.validationService.ValidateKeyPackages(ctx, keyPackageBytes)
+	if err != nil {
+		// TODO: Differentiate between validation errors and internal errors
+		return nil, status.Errorf(codes.InvalidArgument, "invalid identity: %s", err)
+	}
+
+	keyPackageModels := make([]*mlsstore.KeyPackage, len(validationResults))
+	for i, validationResult := range validationResults {
+		kp := mlsstore.NewKeyPackage(validationResult.InstallationId, keyPackageBytes[i], false)
+		keyPackageModels[i] = &kp
```
snormore, line 119: Can `mlsstore.NewKeyPackage` just return a pointer?

pkg/mlsstore/store.go
```diff
@@ -2,27 +2,170 @@ package mlsstore
...
+		return nil, err
 	}
-	s.ctx, s.cancel = context.WithCancel(context.Background())
 
 	return s, nil
 }
+
+func (s *Store) Close() {
+	if s.db != nil {
+		s.db.Close()
```
snormore, line 45: Since `s.db` is instantiated outside and passed in via `Config`, it's probably on the caller to close it down too rather than closing it down in here.

pkg/mlsstore/store.go
```diff
@@ -2,27 +2,170 @@ package mlsstore
...
+	return err
+}
+
+func (s *Store) ConsumeKeyPackages(ctx context.Context, installationIds []string) ([]*KeyPackage, error) {
+	keyPackages := make([]*KeyPackage, 0)
+	err := s.db.RunInTx(ctx, &sql.TxOptions{}, func(ctx context.Context, tx bun.Tx) error {
+		err := tx.NewRaw(`
+			SELECT DISTINCT ON(installation_id) * FROM key_packages
+			WHERE "installation_id" IN (?)
+			AND "consumed_at" IS NULL
```
snormore, line 96: This won't be able to use the `consumed_at` index because it's against null. If we want to optimize this query we can add another `not_consumed` boolean that can be used for these queries, and add an index on `(installation_id, not_consumed, is_last_resort, created_at)`.

pkg/mlsstore/store.go
```diff
@@ -2,27 +2,170 @@ package mlsstore
...
+
+		if len(keyPackages) < len(installationIds) {
+			return errors.New("key packages not found")
+		}
+
+		_, err = tx.NewUpdate().
+			Table("key_packages").
+			Set("consumed_at = ?", nowNs()).
+			Where("is_last_resort = FALSE").
+			Where("id IN (?)", bun.In(extractIds(keyPackages))).
```
snormore, line 114: Want to add the index `(is_last_resort, id)` for this query?

pkg/mlsstore/store_test.go
```diff
@@ -0,0 +1,171 @@
...
+	walletAddress := randomString(32)
+
+	err := store.CreateInstallation(ctx, installationId, walletAddress, randomBytes(32))
+	require.NoError(t, err)
+
+	installationFromDb := &Installation{}
+	require.NoError(t, store.db.NewSelect().Model(installationFromDb).Where("id = ?", installationId).Scan(ctx))
+	require.Equal(t, walletAddress, installationFromDb.WalletAddress)
+
+	keyPackageFromDb := &KeyPackage{}
```
snormore, line 53: Nit: s/Db/DB 🙏

pkg/mlsvalidate/config.go
```diff
@@ -0,0 +1,5 @@
+package mlsvalidate
+
+type MlsValidationOptions struct {
```
snormore, line 3: s/MlsValidationOptions/MLSValidationOptions 🙏

pkg/mlsvalidate/config.go
```diff
@@ -0,0 +1,5 @@
...
+	GrpcAddress string `long:"grpc-address" description:"Address for the GRPC validation service"`
```
snormore, line 4: s/GrpcAddress/GRPCAddress 🙏

pkg/mlsvalidate/service.go
```diff
@@ -0,0 +1,107 @@
...
+type GroupMessageValidationResult struct {
+	GroupId string
+}
+
+type IdentityInput struct {
+	SigningPublicKey []byte
+	Identity         []byte
+}
+
+type MlsValidationService interface {
```
snormore, line 26: s/Mls/MLS 🙏

pkg/server/options.go
```diff
@@ -62,17 +63,19 @@ type Options struct {
...
-	Profiling ProfilingOptions      `group:"DD APM Profiling Options" namespace:"profiling"`
-	MlsStore  mlsstore.StoreOptions `group:"MLS Options" namespace:"mlsstore"`
+	API           api.Options                      `group:"API Options" namespace:"api"`
+	Authz         AuthzOptions                     `group:"Authz Options"`
+	Relay         RelayOptions                     `group:"Relay Options"`
+	Store         store.Options                    `group:"Store Options" namespace:"store"`
+	Metrics       MetricsOptions                   `group:"Metrics Options"`
+	Tracing       TracingOptions                   `group:"DD APM Tracing Options"`
+	Profiling     ProfilingOptions                 `group:"DD APM Profiling Options" namespace:"profiling"`
+	MlsStore      mlsstore.StoreOptions            `group:"MLS Options" namespace:"mls-store"`
```
snormore, line 79: s/Mls/MLS 🙏

pkg/migrations/mls/20231023050806_init-schema.up.sql
```diff
@@ -0,0 +1,44 @@
...
+
+--bun:split
+CREATE INDEX idx_installations_created_at ON installations(created_at);
+
+--bun:split
+CREATE INDEX idx_installations_revoked_at ON installations(revoked_at);
+
+--bun:split
+-- Adding indexes for the key_packages table
+CREATE INDEX idx_key_packages_installation_id ON key_packages(installation_id);
```
snormore, line 35: Most of these individual indexes seem unnecessary based on the queries in `store.go`. I'm only see the need for these 2:
- `(installation_id, not_consumed, is_last_resort, created_at)`
- `(is_last_resort, id)`

pkg/mlsvalidate/service_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	svc "github.com/xmtp/proto/v3/go/mls_validation/v1"
+	"google.golang.org/grpc"
+)
+
+type MockedGrpcService struct {
```
snormore, line 13: s/Grpc/GRPC 🙏

.github/workflows/push-mls.yml
```diff
@@ -0,0 +1,34 @@
+name: Deploy Nodes
+on:
+  push:
+    branches:
+      - mls
```
snormore, line 5: Is the plan to keep this on the `mls` branch for now or merge into `main`? Noticed that the PR is opened against the `mls` branch too

pkg/api/message/v3/service.go
neekolas: 👍🏼

pkg/mlsstore/store.go
neekolas: I was gonna rely on Postgres' ability to combine multiple indexes, but that would work too and would result in smaller index sizes if we aren't using the individual indexes for anything else.

pkg/migrations/mls/20231023050806_init-schema.up.sql
neekolas: There are more methods coming to the store in the next PR that should use some of them. But I do like the idea of adding the `not_consumed` field and a compound index for the most common/expensive query. I'll add that and remove a few of the related indexes.

pkg/mlsstore/store.go
neekolas: I came to the same conclusion and fixed that in the next PR

pkg/api/message/v3/service.go
```diff
@@ -16,21 +17,23 @@ import (
...
 	proto.UnimplementedMlsApiServer
 
-	log          *zap.Logger
-	waku         *wakunode.WakuNode
-	messageStore *store.Store
-	mlsStore     mlsstore.MlsStore
+	log               *zap.Logger
+	waku              *wakunode.WakuNode
+	messageStore      *store.Store
+	MLSStore          mlsstore.MlsStore
```
snormore, line 23: Does this one need to be public or did it get caught up in the find/replace?

pkg/server/options.go
```diff
@@ -62,17 +63,19 @@ type Options struct {
...
-	MlsStore  mlsstore.StoreOptions `group:"MLS Options" namespace:"mlsstore"`
+	API           api.Options                      `group:"API Options" namespace:"api"`
+	Authz         AuthzOptions                     `group:"Authz Options"`
+	Relay         RelayOptions                     `group:"Relay Options"`
+	Store         store.Options                    `group:"Store Options" namespace:"store"`
+	Metrics       MetricsOptions                   `group:"Metrics Options"`
+	Tracing       TracingOptions                   `group:"DD APM Tracing Options"`
+	Profiling     ProfilingOptions                 `group:"DD APM Profiling Options" namespace:"profiling"`
+	MLSStore      mlsstore.StoreOptions            `group:"MLS Options" namespace:"mls-store"`
+	MlsValidation mlsvalidate.MLSValidationOptions `group:"MLS Validation Options" namespace:"mls-validation"`
```
snormore, line 80: s/Mls/MLS here too 🙏

pkg/testing/store.go
```diff
@@ -48,3 +49,17 @@ func NewAuthzDB(t *testing.T) (*bun.DB, string, func()) {
 
 	return bunDB, dsn, cleanup
 }
+
+func NewMlsDB(t *testing.T) (*bun.DB, string, func()) {
```
snormore, line 53: s/Mls/MLS here too 🙏

snormore (approved):

pkg/api/message/v3/service.go
neekolas: Good call. Totally got caught up in the find or replace.

.github/workflows/push-mls.yml
neekolas: Plan is to use the MLS branch for at least this week, and work out all the kinks with the schema. Then merge to main and set the appropriate server flags.

pkg/mlsstore/models.go
```diff
@@ -0,0 +1,24 @@
...
+import "github.com/uptrace/bun"
+
+type Installation struct {
+	bun.BaseModel `bun:"table:installations"`
+
+	ID            string `bun:",pk"`
+	WalletAddress string `bun:"wallet_address,notnull"`
+	CreatedAt     int64  `bun:"created_at,notnull"`
+	RevokedAt     *int64 `bun:"revoked_at"`
+}
```
richardhuaaa, line 12: @neekolas Any reason we're storing the derived state (when an installation was created/revoked) rather than the registration/revocation payloads that were originally used to perform those actions? I think it's important that the client can validate this data themselves

PR #321, opened 2023-10-24
Add more MLS methods

neekolas:
## Summary

- Implements `PublishToGroup` method
- Implements `PublishWelcomes` method
- Implements `GetIdentityUpdates` method
- Adds more validations to incoming RPCs


pkg/store/store.go
```diff
@@ -20,6 +21,22 @@ import (
 
 const maxPageSize = 100
 
+const timestampGeneratorSql = `(
```
neekolas, line 24: I ran all the SQL through a formatter and this is what came out. 🤷

pkg/api/message/v3/service.go
```diff
@@ -134,6 +197,107 @@ func (s *Service) RevokeInstallation(ctx context.Context, req *proto.RevokeInsta
...
+}
+
+func validateGetIdentityUpdatesRequest(req *proto.GetIdentityUpdatesRequest) error {
+	if req == nil || len(req.WalletAddresses) == 0 {
+		return status.Errorf(codes.InvalidArgument, "no wallet addresses to get updates for")
+	}
+	return nil
+}
+
+func isReadyToSend(groupId string, message []byte) error {
```
snormore, line 295: Rename to `requireReadyToSend` or `validateReadyToSend`?

pkg/api/message/v3/service_test.go
```diff
@@ -0,0 +1,309 @@
...
+	v1 "github.com/xmtp/proto/v3/go/message_api/v1"
+	proto "github.com/xmtp/proto/v3/go/message_api/v3"
+	messageContents "github.com/xmtp/proto/v3/go/mls/message_contents"
+	"github.com/xmtp/xmtp-node-go/pkg/mlsstore"
+	"github.com/xmtp/xmtp-node-go/pkg/mlsvalidate"
+	"github.com/xmtp/xmtp-node-go/pkg/store"
+	test "github.com/xmtp/xmtp-node-go/pkg/testing"
+)
+
+type mockedMlsValidationService struct {
```
snormore, line 21: Nit: s/Mls/MLS 😅

pkg/migrations/mls/20231023050806_init-schema.up.sql
```diff
@@ -33,7 +33,7 @@ CREATE INDEX idx_installations_revoked_at ON installations(revoked_at);
 
 --bun:split
 -- Adding indexes for the key_packages table
-CREATE INDEX idx_key_packages_installation_id_not_is_last_resort_created_at ON key_packages(
+CREATE INDEX idx_key_packages_installation_id_not_consumed_is_last_resort_created_at ON key_packages(
```
snormore, line 36: Surprised this doesn't hit the index name length limit

pkg/mlsstore/store.go
```diff
@@ -125,6 +121,59 @@ func (s *Store) ConsumeKeyPackages(ctx context.Context, installationIds []string
...
+
+	return out, nil
+}
+
+func (s *Store) RevokeInstallation(ctx context.Context, installationId string) error {
+	_, err := s.db.NewUpdate().
+		Model(&Installation{}).
+		Set("revoked_at = ?", nowNs()).
+		Where("id = ?", installationId).
+		Where("revoked_at IS NULL").
```
snormore, line 171: Could make sense to use `not_revoked` here too

pkg/store/store.go
```diff
@@ -127,6 +144,61 @@ func (s *Store) InsertMessage(env *messagev1.Envelope) (bool, error) {
 	return stored, err
 }
 
+func (s *Store) InsertMlsMessage(ctx context.Context, contentTopic string, data []byte) (*messagev1.Envelope, error) {
```
snormore, line 147: s/Mls/MLS 🙏

pkg/mlsstore/store.go
snormore: Since it's scoping on installation ID first it's probably not a big deal to just use revoked_at IS NULL tbh

snormore (approved):

pkg/mlsstore/store.go
```diff
@@ -125,6 +121,59 @@ func (s *Store) ConsumeKeyPackages(ctx context.Context, installationIds []string
...
+	updated := make([]*Installation, 0)
+	// Find all installations that were changed since the startTimeNs
+	err := s.db.NewSelect().
+		Model(&updated).
+		Where("wallet_address IN (?)", bun.In(walletAddresses)).
+		WhereGroup(" AND ", func(q *bun.SelectQuery) *bun.SelectQuery {
+			return q.Where("created_at > ?", startTimeNs).WhereOr("revoked_at > ?", startTimeNs)
+		}).
+		Order("created_at ASC").
+		Scan(ctx)
```
richardhuaaa, line 134: If I understood correctly, the server just returns a list of installations with their status (registered/revoked)? And we're not attaching the original signed payloads?

So the issue with this approach is we're relying 100% on server trust, not just for ordering but for determining who to encrypt messages for. The server can add and remove arbitrary clients for any wallet address, which circumvents the E2EE.

neekolas: The server can definitely hide arbitrary installations. Adding a fake installation will not get the server very far. The purpose of this endpoint is to get a concise list of `installation_id`s which the client may then request key packages for. If the key package doesn't exist, or doesn't validate correctly, it would be ignored and could be flagged to the user.

Revocation isn't fully fleshed out. We are going to do want to include proof for that once we figure out what the revocation payloads actually look like. The protos are setup to allow us to add more fields to the revocation updates once we know what to put in there.

richardhuaaa: I see, so this can work. My next question here is about overall design. In order to verify the server's result here, the client MUST then fetch the corresponding `KeyPackages`. Essentially, the current design assumes that we will never want to fetch a `Credential` without fetching a `KeyPackage`.

 Here are the types of query patterns that I anticipate we will need:
 
 1. What are my active installations, so that I can revoke them if needed?
 2. What are the other users' active installations in this conversation, so that I can verify the membership is correct?
 3. I want to start a conversation with another user - give me all of their key packages.
 
I believe we only want KeyPackages for (3), and for (1) and (2) we don't necessarily want to consume any one-time keypackages. We could issue the query for LastResortKeyPackages specifically, but that seems inelegant. Naively, I would think (1) and (2) could be cleaner by fetching Credentials and Revocations directly, rather than an unsigned list of installations provided by the server?

neekolas: We synced offline. I am going to update things to include both the credential and public key on the installation, so that we can include it in `GetIdentityUpdates` for client verification.

neekolas:
I've switched things in both the Rust validation service and the node so that `installation_id` is just the full public key of the installation as bytes. I'm also now storing the credential in the DB and including it in the `GetIdentityUpdates` response, so that clients can perform validation purely on the response from `GetIdentityUpdates`.

PR #326, opened 2023-12-29
Emit subscribed topics metric

snormore:
Emit metric that indicates number of subscribed topics per request.


neekolas (approved):
Looks good to me

PR #329, opened 2024-01-05
Assume KeyPackages are always last resort.

Bren2010:


go.mod
```diff
@@ -187,3 +187,5 @@ require (
 
 // From node-go
 replace github.com/ethereum/go-ethereum v1.10.17 => github.com/status-im/go-ethereum v1.10.4-status.2
+
+replace github.com/xmtp/proto/v3 v3.36.1-0.20231219054634-2ff03b7d5090 => github.com/Bren2010/proto/v3 v3.0.0-20240104022019-f50c3c99cd1d
```
neekolas, line 191: Let's get this into `xmtp/proto` main before merging

pkg/mlsstore/store.go
```diff
@@ -43,83 +39,57 @@ func New(ctx context.Context, config Config) (*Store, error) {
...
+	}
 
-		_, err = tx.NewUpdate().
-			Table("key_packages").
-			Set("consumed_at = ?", nowNs()).
-			Set("not_consumed = FALSE").
-			Where("is_last_resort = FALSE").
-			Where("id IN (?)", bun.In(extractIds(keyPackages))).
-			Exec(ctx)
+	_, err := s.db.NewUpdate().
```
neekolas, line 73: I don't believe this will return an error if no installations matched the update query. We might want to change this to error out in that case, since the caller would want to know that their request did nothing.

neekolas (approved):

pkg/mlsstore/store.go
Bren2010: This change was documented in the proto PR: https://github.com/xmtp/proto/pull/123/files#diff-89ac5361df54d78c25ce625d56f63dd01e1bd813456a5430d992ff33c8fc3a2dR143

It seemed brittle to have the whole request fail if one installation doesn't exist. I can remove if it would be an issue though

Bren2010: Nevermind, I thought comment this was on FetchKeyPackages. Let me try to fix

Issue #319, opened 2023-10-20
Feature request: Total ordering of V3 messages


neekolas:
### Is your feature request related to a problem?

The initial approach we were planning to take with MLS messages was to keep track of group epochs in a database table and reject messages sent from older epochs. This is similar to the [hacky example delivery service](https://github.com/openmls/openmls/blob/main/delivery-service/ds/src/main.rs#L290) in the OpenMLS repo.

Because the nodes have a limited ability to validate messages (everything but the group ID and epoch are encrypted), this poses an issue. An invalid update could increment the group epoch (even from someone outside the group), causing otherwise valid messages to fail validation.

@bren2010 proposed that we just enforce total ordering of messages on the nodes and keep the nodes unaware of epochs. Clients would be responsible for discarding messages that come from conflicting epochs, and re-sending updates if a previous update ends up being discarded.

### Describe the solution to the problem

Even in our current centralized current system, total ordering doesn't come for free. We have multiple nodes with clocks that can drift out of sync, so simply sequencing messages based on server timestamps will fail in some small percentage of cases due to race conditions and clock conflicts.

The goal should be that the client can completely trust that when they receive the latest messages from the nodes, no earlier messages will ever appear in the topic.

My proposal to solve this is pretty simple. When publishing a MLS message, we rely on the database's NOW() function to set the `sender_timestamp` field in the DB (for MLS messages only). The `sender_timestamp` is what is already used for sort in our `Query` and `BatchQuery` APIs.

It's not 100% fool-proof - you can still get conflicts at the microsecond level, and during a database failover the clock could jump backwards between nodes - but it should be a strong enough guarantee we can build systems that rely on the ordering.

As we decentralize the network we will need to develop more advanced conflict resolution strategies, since this kind of total ordering is not possible in a decentralized network. But we can defer that work until closer to the time of decentralization.

### Additional details

You have to do a little bit of gymnastics to get postgres to give you a nanosecond timestamp (what we already have in our schema). And the best it can give you is microsecond precision multiplied by 1000.

Something like this should work:

```sql
((EXTRACT(EPOCH FROM clock_timestamp())::bigint * 1000000000) 
       + (EXTRACT(MICROSECONDS FROM clock_timestamp())::bigint * 1000));
```

There are fancier methods (logical clocks, serializable transactions that check for newer messages and roll back, using sequences as part of the sort), but I think this should do the job and has a minimal performance penalty.

richardhuaaa:
[...]

PR #330, opened 2024-01-10
Store MLS messages separately

snormore:

What this includes:
- Add messages table to MLS DB
- Implement InsertMessage and QueryMessages in MLS store

What this doesn't include that the existing messages DB has:
- Cleaner process for MLS messages
- DB metrics for MLS messages
- Reader connection for MLS DB
- E2E tests for MLS topics that write to this DB

### DIFF
--- a/pkg/api/message/v1/service.go
+++ b/pkg/api/message/v1/service.go
@@ -20,6 +20,7 @@ import (
 	apicontext "github.com/xmtp/xmtp-node-go/pkg/api/message/v1/context"
 	"github.com/xmtp/xmtp-node-go/pkg/logging"
 	"github.com/xmtp/xmtp-node-go/pkg/metrics"
+	"github.com/xmtp/xmtp-node-go/pkg/mlsstore"
 	"github.com/xmtp/xmtp-node-go/pkg/store"
 	"github.com/xmtp/xmtp-node-go/pkg/topic"
 	"github.com/xmtp/xmtp-node-go/pkg/tracing"
@@ -44,9 +45,10 @@ type Service struct {
 	proto.UnimplementedMessageApiServer
 
 	// Configured as constructor options.
-	log   *zap.Logger
-	waku  *wakunode.WakuNode
-	store *store.Store
+	log      *zap.Logger
+	waku     *wakunode.WakuNode
+	store    *store.Store
+	mlsStore mlsstore.MlsStore
 
 	// Configured internally.
 	ctx       context.Context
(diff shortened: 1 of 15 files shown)


pkg/mlsstore/store.go
```diff
@@ -149,14 +161,144 @@ func (s *Store) GetIdentityUpdates(ctx context.Context, walletAddresses []string
...
+	}
+
+	return &messagev1.Envelope{
+		ContentTopic: topic,
+		TimestampNs:  uint64(message.CreatedAt),
+		Message:      content,
+	}, nil
+}
+
+func (s *Store) QueryMessages(ctx context.Context, query *messagev1.QueryRequest) (*messagev1.QueryResponse, error) {
```
snormore, line 197: The mixing of messagev1 types into this feels a bit uncomfortable

pkg/api/message/v1/service.go
```diff
@@ -367,6 +370,24 @@ func (s *Service) Query(ctx context.Context, req *proto.QueryRequest) (*proto.Qu
...
+		}
+	}
+	if hasV3Topic && hasNonV3Topic {
+		return nil, errors.New("cannot query v3 topics with non-v3 topics")
+	}
+	if hasV3Topic {
+		if s.mlsStore == nil {
+			return nil, errors.New("mls store not configured")
+		}
+		return s.mlsStore.QueryMessages(ctx, req)
```
snormore, line 388: Alternatively we could add `QueryMessages` to the mls/v3 service with it's own isolated types, rather than hijacking the messagev1 service+types for it

neekolas: Downstream impact isn't a huge deal right now. There isn't that much downstream at the moment. Basically a single caller.

snormore: Ok great I think I'll add the mls/v3 protos to separate this all out cleanly then

snormore:
Converted back into a draft PR while I do the separated proto work for querying.

pkg/mlsstore/store.go
```diff
@@ -149,14 +161,144 @@ func (s *Store) GetIdentityUpdates(ctx context.Context, walletAddresses []string
...
+		Set("revoked_at = ?", s.nowNs()).
 		Where("id = ?", installationId).
 		Where("revoked_at IS NULL").
 		Exec(ctx)
 
 	return err
 }
 
+func (s *Store) InsertMessage(ctx context.Context, topic string, content []byte) (*messagev1.Envelope, error) {
+	createdAt := s.config.now()
```
neekolas, line 173: I'm not sure this ordering is as total as we want. Clock drift between nodes could have newer messages actually end up before older messages.

pkg/migrations/mls/20240109001927_add-messages.up.sql
```diff
@@ -0,0 +1,15 @@
+SET statement_timeout = 0;
+
+--bun:split
+
+CREATE TABLE messages (
+    topic VARCHAR(200) NOT NULL,
+    tid VARCHAR(96) NOT NULL,
```
neekolas, line 7: Given how strict the ordering requirements are, I'd really rather this rely on a sequence than the server time. Clocks can drift, and any re-ordering can corrupt the group state with the way MLS works. Even if those conflicts are improbable, it makes debugging a lot easier if they are impossible.

snormore: Ok agree, that makes sense. I'll update it to do that. 

I'm going to close down this PR as it is, I think I'd also like to just split up the messages by type too; group vs welcome messages, and get rid of the idea of topics completely.

PR #332, opened 2024-01-11
refactor: move api/message/v3 to mls/api/v1

snormore:
PR to match the changes in https://github.com/xmtp/proto/pull/126 - Pulling some individual changes out of a bigger update I've been working on to decouple MLS querying from messagev1.

This PR also moves pkg/mlsstore to pkg/mls/store, along with pkg/mls/api since this code will eventually be completely isolated anyway so no need to mix it in with the other pkgs.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,7 +30,7 @@ require (
 	github.com/uptrace/bun/driver/pgdriver v1.1.16
 	github.com/waku-org/go-waku v0.8.0
 	github.com/xmtp/go-msgio v0.2.1-0.20220510223757-25a701b79cd3
-	github.com/xmtp/proto/v3 v3.36.1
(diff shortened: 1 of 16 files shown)


go.mod
```diff
@@ -30,7 +30,7 @@ require (
 	github.com/uptrace/bun/driver/pgdriver v1.1.16
 	github.com/waku-org/go-waku v0.8.0
 	github.com/xmtp/go-msgio v0.2.1-0.20220510223757-25a701b79cd3
-	github.com/xmtp/proto/v3 v3.36.1
+	github.com/xmtp/proto/v3 v3.36.3-0.20240111132545-4e1d2b1b2399
```
snormore, line 33: This will be updated to the proper non-sha version after https://github.com/xmtp/proto/pull/126 merges and is released

neekolas (approved):

PR #333, opened 2024-01-11
Decouple MLS messages from messagev1

snormore:
- Add group_messages and welcome_messages table to MLS DB
- Implement InsertGroupMessages, InsertWelcomeMessages, QueryGroupMessages, and QueryWelcomeMessages in MLS store
- Implement QueryGroupMessages and QueryWelcomeMessages in MLS service
- Implement SubscribeGroupMessages and SubscribeWelcomeMessages in MLS service


pkg/migrations/mls/20240109001927_add-messages.up.sql
```diff
@@ -0,0 +1,27 @@
+SET statement_timeout = 0;
+
+--bun:split
+
+CREATE TABLE group_messages (
+    id BIGSERIAL PRIMARY KEY,
+    created_at TIMESTAMP NOT NULL DEFAULT NOW(),
+    group_id VARCHAR(200) NOT NULL,
+    data BYTEA
```
neekolas, line 9: One thing we are going to want on this API is idempotency, like we have on the V1 publish APIs. If we had a unique index on a field that represented `HASH(CONCAT(group_id, data))` we could safely retry requests without worrying about duplicate messages.

pkg/migrations/mls/20240109001927_add-messages.up.sql
```diff
@@ -0,0 +1,27 @@
+SET statement_timeout = 0;
+
+--bun:split
+
+CREATE TABLE group_messages (
+    id BIGSERIAL PRIMARY KEY,
+    created_at TIMESTAMP NOT NULL DEFAULT NOW(),
+    group_id VARCHAR(200) NOT NULL,
```
neekolas, line 8: Think this is fine to store as text, but at the `libxmtp` and `openmls` level `group_id`s are always byte arrays. We can stringify in the node, obviously, but it might be simpler to just keep them as byte arrays all the way.

pkg/migrations/mls/20240109001927_add-messages.up.sql
```diff
@@ -0,0 +1,27 @@
...
+
+CREATE INDEX idx_group_messages_group_id_created_at ON group_messages(group_id, created_at);
+
+--bun:split
+
+CREATE TABLE welcome_messages (
+    id BIGSERIAL PRIMARY KEY,
+    created_at TIMESTAMP NOT NULL DEFAULT NOW(),
+    installation_id VARCHAR(200) NOT NULL,
+    data BYTEA
```
neekolas, line 22: Same question around idempotency as above

pkg/migrations/mls/20240109001927_add-messages.up.sql
snormore: We actually have (topic, timestamp, content) uniqueness in v1, not (topic, content), so you can technically send the same content at a different time and it wouldn't be a duplicate. If there's no timestamp embedded in `data` then it would mean that you can't send the same message more than once if we do (group_id, data).

pkg/migrations/mls/20240109001927_add-messages.up.sql
snormore: Installation ID too or just group ID?

neekolas: They're both byte arrays from the client's perspective. The installation id is actually a public key.

pkg/migrations/mls/20240109001927_add-messages.up.sql
neekolas: You really shouldn't be able to construct two identical messages with MLS. There's always nonces and salts somewhere in there, changing up the encryption.

dev/migrate-mls
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -e
+
+dev/run --create-mls-migration "$@"
```
neekolas, line 4: Thanks for adding this. Totally slipped my mind when I was setting up the new DB

pkg/mls/api/v1/service.go
```diff
@@ -174,8 +174,8 @@ func (s *Service) SendGroupMessages(ctx context.Context, req *proto.SendGroupMes
 		}
 
 		// TODO: Wrap this in a transaction so publishing is all or nothing
-		wakuTopic := topic.BuildGroupTopic(result.GroupId)
-		msg, err := s.store.InsertGroupMessage(ctx, result.GroupId, data)
+		wakuTopic := topic.BuildGroupTopic([]byte(result.GroupId))
+		msg, err := s.store.InsertGroupMessage(ctx, []byte(result.GroupId), data)
```
neekolas, line 178: I actually hex encode the group_id in the validation service. I should probably unwind that decision and just make it return the bytes, but until then you can hex decode it here to get the original value. When we built the service we thought we'd want to treat group_ids as strings.

I'll take care of making a PR to change the validation service return type to bytes once all of this has landed.

pkg/migrations/mls/20240109001927_add-messages.up.sql
```diff
@@ -0,0 +1,37 @@
...
+--bun:split
+
+CREATE UNIQUE INDEX idx_group_messages_group_id_data_hash ON group_messages (group_id_data_hash);
+
+--bun:split
+
+CREATE TABLE welcome_messages (
+    id BIGSERIAL PRIMARY KEY,
+    created_at TIMESTAMP NOT NULL DEFAULT NOW(),
+    installation_id BYTEA NOT NULL,
```
richardhuaaa, line 26: If we're using raw bytes here it might be better to call it `installation_key` or `installation_public_key` - when shown in UX, we will stringify it and call it `installation_id`: https://github.com/xmtp/libxmtp/blob/main/xmtp_mls/IDENTITY.md#installation-registration

snormore: Makes sense - rename all the `bytes installation_id` references on the protos to `installation_public_key` too?

pkg/api/interceptor.go
```diff
@@ -87,23 +87,9 @@ func (wa *WalletAuthorizer) Stream() grpc.StreamServerInterceptor {
...
-		if !strings.HasPrefix(envelope.ContentTopic, "/xmtp/mls/") {
-			return false
-		}
-	}
-	return true
-}
-
 func (wa *WalletAuthorizer) requiresAuthorization(req interface{}) bool {
-	publishRequest, isPublish := req.(*messagev1.PublishRequest)
-	return isPublish && (!wa.isProtocolMLS(publishRequest) || wa.AuthnConfig.EnableMLS)
```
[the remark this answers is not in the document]
neekolas: I believe this interceptor no longer runs at all for MLS publishing requests, because we are using special new Send* methods.

pkg/migrations/mls/20240109001927_add-messages.up.sql
```diff
@@ -0,0 +1,37 @@
+SET statement_timeout = 0;
+
+--bun:split
+
+CREATE TABLE group_messages (
+    id BIGSERIAL PRIMARY KEY,
+    created_at TIMESTAMP NOT NULL DEFAULT NOW(),
```
neekolas, line 7: Good call with the regular timestamp. Totally fine to not have nanosecond precision here. We aren't even using this for sorting anymore, the timestamps are purely informational.

pkg/mls/store/store.go
```diff
@@ -157,8 +171,189 @@ func (s *Store) RevokeInstallation(ctx context.Context, installationId []byte) e
...
+					CreatedNs: uint64(msg.CreatedAt.UnixNano()),
+					GroupId:   msg.GroupId,
+					Data:      msg.Data,
+				},
+			},
+		})
+	}
+
+	pagingInfo := &mlsv1.PagingInfo{Limit: 0, IdCursor: 0, Direction: direction}
+	if len(messages) >= pageSize {
```
neekolas, line 274: Might be more intuitive to write this as `if len(messages >= pageSize && pageSize > 0 {`, which is really what we're doing here.

pkg/mls/store/store.go
```diff
@@ -157,8 +171,189 @@ func (s *Store) RevokeInstallation(ctx context.Context, installationId []byte) e
...
+					Id:        msg.Id,
+					CreatedNs: uint64(msg.CreatedAt.UnixNano()),
+					GroupId:   msg.GroupId,
+					Data:      msg.Data,
+				},
+			},
+		})
+	}
+
+	pagingInfo := &mlsv1.PagingInfo{Limit: 0, IdCursor: 0, Direction: direction}
```
neekolas, line 273: We should use the original limit, so that the caller can just pass back the `PagingInfo` into the next request and not be surprised by the number of results.

pkg/mls/store/store.go
```diff
@@ -157,8 +171,189 @@ func (s *Store) RevokeInstallation(ctx context.Context, installationId []byte) e
...
+				V1: &mlsv1.WelcomeMessage_V1{
+					Id:        msg.Id,
+					CreatedNs: uint64(msg.CreatedAt.UnixNano()),
+					Data:      msg.Data,
+				},
+			},
+		})
+	}
+
+	pagingInfo := &mlsv1.PagingInfo{Limit: 0, IdCursor: 0, Direction: direction}
```
neekolas, line 341: Same as above for re-using the original `pageSize`

pkg/mls/store/store.go
neekolas: Actually, I guess `pageSize` can never be 0 in practice. If you specify that you'll get `maxPageSize`. So, I'm not sure the if statement below does anything.

.github/workflows/push-mls.yml
```diff
@@ -28,7 +28,7 @@ jobs:
       - name: Push
         id: push
         run: |
-          export DOCKER_IMAGE_TAG=mls
+          export DOCKER_IMAGE_TAG=mls-dev
```
snormore, line 31: TODO: revert this commit before merging

pkg/mls/api/v1/service.go
```diff
@@ -77,220 +157,354 @@ func (s *Service) FetchKeyPackages(ctx context.Context, req *proto.FetchKeyPacka
...
+		}
 
-func (s *Service) PublishWelcomes(ctx context.Context, req *proto.PublishWelcomesRequest) (res *emptypb.Empty, err error) {
-	if err = validatePublishWelcomesRequest(req); err != nil {
-		return nil, err
-	}
+		// TODO: Wrap this in a transaction so publishing is all or nothing
+		decodedGroupId, err := hex.DecodeString(result.GroupId)
+		if err != nil {
+			return nil, err
```
neekolas, line 254: We can add it in a subsequent PR, but would be nice to have these return GRPC status codes to the client. The raw errors will all just turn into `INTERNAL`

neekolas (approved):

PR #335, opened 2024-01-19
Generate protos in repo

neekolas:
## Summary

I'm trying to clean up `xmtp/proto`. I'd like to remove the Go artifacts and just have them get generated in the consuming repo. This does that using Buf hosted plugins. All the code generation happens remotely, so there is no special configuration you need to do locally beyond having the Buf CLI setup.


buf.gen.yaml
```diff
@@ -0,0 +1,28 @@
+# buf.gen.yaml
```
neekolas, line 1: This is where all the configuration lives

PR #336, opened 2024-01-20
MLS subscribe via db and optionally from a cursor

snormore:

- Update MLS subscribe group/welcome messages service methods to poll the DB for streamed messages rather than relying on libp2p-pubsub, so that ordering is always consistent.
- Implements logic for subscribing from a specific cursor so that clients can resume subscriptions without losing track of messages in between reconnects.


.github/workflows/push-mls.yml
```diff
@@ -2,7 +2,7 @@ name: Push MLS Container
 on:
   push:
     branches:
-      - mls
+      - snor/mls-msgs
```
snormore, line 5: Will be reverted before merging this PR

.github/workflows/push-mls.yml
```diff
@@ -2,7 +2,7 @@ name: Push MLS Container
 on:
   push:
     branches:
-      - mls
+      - snor/mls-subscribe
```
snormore, line 5: TODO: revert before merging this PR

neekolas:
So, I just want to do some back of the envelope math here.

If a user is subscribed to 100 groups with no cursor, this will cause:
- 100 DB queries at the start of the subscription to get the last cursor
- 20 DB queries per second because of the `passiveTicker`

With a cursor it would just be the 20/second.

Is that right?

I don't really know what our scaling limits are. Maybe this is fine with some appropriate rate limits. But we do get about 1.8M subscribe requests a day, some of them lasting hours, so we should proceed cautiously here. I can put up a PR to collect metrics on how many concurrent subscriptions we handle today.

There is some low hanging fruit to optimize. We could collapse the 20 queries/sec into a single more complex query with each topic/cursor pair OR'd together. Might help a bit. 

We could also be running these queries against the read replica. Delays on the order of replica lag are fine, although it does add the possibility that the nats message will make it to the subscriber faster than the database change and the query will return nothing.

pkg/mls/api/v1/service.go
```diff
@@ -88,28 +87,18 @@ func (s *Service) Close() {
...
-			return err
-		}
-		if msg.GetV1() == nil {
-			return nil
-		}
-		err = s.nc.Publish(buildNatsSubjectForWelcomeMessages(msg.GetV1().InstallationKey), wakuMsg.Payload)
+		// The waku message payload is just the installation key as bytes since
+		// we only need to use it as a signal that a new message was published,
+		// without any other content.
+		err := s.nc.Publish(buildNatsSubjectForWelcomeMessages(wakuMsg.Payload), wakuMsg.Payload)
```
neekolas, line 101: One thing that is a hard requirement here is some way of supporting `streamAllMessages`, since we know integrators are going to want to use their push servers with MLS. Is the idea that each `streamAllMessages` connection would poll the database for any new messages that have arrived since the last time it checked?

snormore:
Would be the same number of queries with and without a cursor in the request; that shouldn't affect the rate of queries, just where it starts querying from.

In your example of a request with 100 groups/filters, it would be 100 queries to get the latest cursor, 1 for each, and then a query every 5 seconds because of the passive ticker assuming no other messages are sent to those groups. We can always decrease the passive ticker period, that's really just there in case a pubsub message doesn't get delivered, which shouldn't happen at all or very rarely if it does.

We could combine the initial start-cursor queries into 1 if asked for many groups, but they should be very quick without putting much pressure on the DBs as long as we have the right index, so I don't know if I'm too worried about those, and we do them all in parallel anyway.

Similarly, the message queries for a batch of groups could be combined, but they should be very quick individually without causing the DBs to do a lot of work too if we have the right indexes, and when done individually they're all done in parallel anyway with the connection pool rate limiting behind the scenes. I don't know if I'm too worried about those either tbh, but we'd keep an eye on telemetry to build confidence too.

Using a read replica makes sense if we need to, that seems reasonable.

I think the DB will handle it without significant effort tbh. It's adding a bunch of queries, but they're all indexed queries that should be mostly just O(1).

pkg/mls/api/v1/service.go
snormore: Yep SubscribeAll logic would look similar but without the group/installation scoping. It could use the nats wildcard topic as signal for activity, but not sure if we'd really need the signal in the all case vs just assuming there's always activity. I think it'd probably be reasonable to have a slightly larger ticker period for SubscribeAll too if we want to tune it a bit for that case

snormore:
I think there's a case to be made that we could rename the Subscribe methods to `GetGroupMessages` and `GetWelcomeMessages` and remove the Query methods completely in favor of just using these. That would probably also result in most requests having an initial cursor included when libxmtp is catching up / syncing for new messages since last checked.

neekolas:
> I think the DB will handle it without significant effort tbh. They're all indexed queries that shouldn't cause any scans or temp table usage, so should be very easy for the DB to process.

Idk. You're right that the queries are small, but we're talking about substantial numbers here. We have a single bot with 150k conversations ongoing. As that moves to V3 that _one wallet_ would be generating 30K QPS. Even Redis starts to run into trouble once you get past 100k QPS.

neekolas:
> There's probably a case to be made to rename the Subscribe methods to GetGroupMessages and GetWelcomeMessages and remove the Query methods completely in favor of just using these. It's a simpler interface, and most requests would be including an initial cursor when libxmtp is catching up / syncing from the last time it checked.

I don't think it really lines up with the ways we currently use the Query APIs in `libxmtp`. A lot of the time we are just trying to synchronize our local state before we do something else, so we really want the results to end when you hit the newest message.

snormore:
> Idk. You're right that the queries are small, but we're talking about substantial numbers here. We have a single bot with 150k conversations ongoing. As that moves to V3 that one wallet would be generating 30K QPS. Even Redis starts to run into trouble once you get past 100k QPS.

150k conversations ongoing doesn't necessarily result in that many queries though:
- There will be an initial number of queries for the start cursor, and I'm not even sure if we'll have those if libxmtp is including it's own cursor for syncing on it's requests.
- From there it depends on how active the conversations are. Queries happen when there is activity, or via the passive ticker if there is no activity at all. In the worst case of all conversations having non-stop activity you do have a lot of queries, but they're going to be rate limited by the DB connection pool anyway.

snormore:
> I don't think it really lines up with the ways we currently use the Query APIs in libxmtp. A lot of the time we are just trying to synchronize our local state before we do something else, so we really want the results to end when you hit the newest message.

The get messages request could have a config for follow/no-follow so it can still behave like query if it's needed.

neekolas:
> The get messages request could have a config for follow/no-follow so it can still behave like query if it's needed. The results just get streamed instead of going through many requests using pagination.

It's all possible. But there are reasons most companies don't just switch all their APIs to streaming. With no upper bound on the size of a response, metrics get harder to track. Rate limiting is harder. Errors and retries are non-standard. I'm perfectly fine with our boring query APIs.

snormore:
Sure, but we already have the streaming equivalent that people can/will use, and that we'll use ourselves for some things, so we're not avoiding the complexity of streaming that you're describing.

neekolas:
I don't see any harm merging this as-is. It looks like everything works, and we'll learn more having it deployed.

The thing I'm apprehensive about is modifying the client to assume that these streaming APIs have the properties of completeness and total ordering. Once production clients start shipping with that assumption baked in, it's very hard to roll back. 

So before we take the time to rewrite a bunch of client code to take full advantage of this more powerful API I'd want to be very confident that this is going to be sustainable for a world where 100% of our network traffic is running on it. It would be a real pain to have to migrate back.

snormore:
Discussed with Nick IRL, and going to close this in favor of just using query in the client when necessary, since consistent ordering is only needed for a subset/type of messages that the client knows about post-decryption.

PR #337, opened 2024-01-22
Add hpke public key

neekolas:
## Summary

To encrypt welcome messages we need to add the `hpke_public_key` field to the stored welcomes and return it in the query results.


.github/workflows/push-mls.yml
```diff
@@ -2,7 +2,8 @@ name: Push MLS Container
 on:
   push:
     branches:
-      - mls
```
neekolas, line 5: Will reverse this before merging

pkg/mls/store/models.go
```diff
@@ -36,4 +36,5 @@ type WelcomeMessage struct {
 	CreatedAt       time.Time `bun:",notnull"`
 	InstallationKey []byte    `bun:",notnull,type:bytea"`
 	Data            []byte    `bun:",notnull,type:bytea"`
+	HpkePublicKey   []byte    `bun:"hpke_public_key,notnull,type:bytea"`
```
snormore, line 39: Nit: s/Hpke/HPKE

snormore (approved):

pkg/mls/store/models.go
neekolas: It might add more confusion than it's worth. The generated protos are always going to be `Hpke`, so changing the model would give the DB the same field name with different casing.

PR #339, opened 2024-01-23
Add subscribed topics gauge metric

snormore:
Metric for better tracking subscribes and unsubscribes.


neekolas (approved):

pkg/api/message/v1/service.go
```diff
@@ -217,11 +217,15 @@ func (s *Service) Subscribe(req *proto.SubscribeRequest, stream proto.MessageApi
...
 		log.Debug("stream closed")
-		return nil
+		break
 	case <-s.ctx.Done():
 		log.Info("service closed")
-		return nil
+		break
 	}
+
+	metrics.EmitUnsubscribeTopics(stream.Context(), log, len(req.ContentTopics))
```
neekolas, line 226: Don't we want this in a `defer`, actually. Otherwise an error will increment the subscription but not the unsubscribe.

Issue #344, opened 2024-01-23
Feature request: MLS subscribe from cursor


snormore:
### Is your feature request related to a problem?

When subscribing for welcome or group messages, a user may disconnect and reconnect, and needs to be able to do so reliably without missing messages in-between.

### Describe the solution to the problem

Add support for subscribing from a given cursor, so that messages that have already been stored from that cursor are streamed, along with any newly arriving messages as usual.

### Additional details

The proto changes have already been added for this https://github.com/xmtp/proto/blob/main/proto/mls/api/v1/mls.proto#L320

PR #351, opened 2024-02-01
MLS subscribe from cursor

snormore:
Adds support to the MLS subscribe {group,welcome} messages endpoints for resuming from a given cursor id.

Fixes https://github.com/xmtp/xmtp-node-go/issues/344

### DIFF
--- a/pkg/mls/api/v1/service.go
+++ b/pkg/mls/api/v1/service.go
@@ -362,8 +362,32 @@ func (s *Service) SubscribeGroupMessages(req *mlsv1.SubscribeGroupMessagesReques
 	// See: https://github.com/xmtp/libxmtp/pull/58
 	_ = stream.SendHeader(metadata.Pairs("subscribed", "true"))
 
-	var streamLock sync.Mutex
+	streamed := map[string]*mlsv1.GroupMessage{}
+	var streamingLock sync.Mutex
+	streamMessages := func(msgs []*mlsv1.GroupMessage) {
+		streamingLock.Lock()
+		defer streamingLock.Unlock()
+
+		for _, msg := range msgs {
+			if msg.GetV1() == nil {
+				continue
+			}
+			encodedId := fmt.Sprintf("%x", msg.GetV1().Id)
+			if _, ok := streamed[encodedId]; ok {
+				log.Debug("skipping already streamed message", zap.String("id", encodedId))
+				continue
+			}
+			err := stream.Send(msg)
+			if err != nil {
+				log.Error("error streaming group message", zap.Error(err))
+			}
+			streamed[encodedId] = msg
+		}
+	}
+
 	for _, filter := range req.Filters {
+		filter := filter
+
 		natsSubject := buildNatsSubjectForGroupMessages(filter.GroupId)
 		log.Info("subscribing to nats subject", zap.String("subject", natsSubject), zap.String("group_id", hex.EncodeToString(filter.GroupId)))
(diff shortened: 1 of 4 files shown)


neekolas:
Do you mind holding off merging until next week. I want to make sure the clients are prepared for potentially old messages to come back. Right now, I think reconnects are going to lead to a lot of dupe messages.

richardhuaaa:
@neekolas @snormore is this ready to merge now?

neekolas:
I think the only gotcha is going to be that we aren't updating the cursor when streamed messages come in (except for commits). So, calling this multiple times will lead to dupe messages.

richardhuaaa:
Gotcha. How about an in-memory cursor on the client? If that makes sense I can get that together and then we can land this one.

neekolas:
I think that's fine. We could also use the last message in the db for streaming purposes, since we aren't shooting for total exhaustiveness.

pkg/mls/api/v1/service.go
```diff
@@ -345,8 +345,32 @@ func (s *Service) SubscribeGroupMessages(req *mlsv1.SubscribeGroupMessagesReques
...
+			err := stream.Send(msg)
+			if err != nil {
+				log.Error("error streaming group message", zap.Error(err))
+			}
+			streamed[encodedId] = msg
+		}
+	}
+
 	for _, filter := range req.Filters {
+		filter := filter
```
richardhuaaa, line 372: Why is this assignment necessary?

pkg/mls/api/v1/service.go
```diff
@@ -371,6 +388,43 @@ func (s *Service) SubscribeGroupMessages(req *mlsv1.SubscribeGroupMessagesReques
 		defer func() {
 			_ = sub.Unsubscribe()
 		}()
+
+		if filter.IdCursor > 0 {
```
richardhuaaa, line 392: Hopefully I'm understanding this correctly - this is happening sequentially after the streaming from the relay is set up, so some of the newer messages from the relay might come in before we deliver these older messages?

I guess the reason for this ordering is to avoid gaps? And it might be possible to receive the messages from the relay, but defer pushing them until after the sync from DB is complete, but it's unnecessarily complex?

snormore: Yep exactly. This is just the simplest approach, but we could defer the relayed messages with the tradeoff of complexity and resource/memory usage.

pkg/mls/api/v1/service.go
snormore: This [is needed](https://eli.thegreenplace.net/2019/go-internals-capturing-loop-variables-in-closures/) because `filter` is used within the goroutine function below. Alternatively you can have the function accept it as an arg. This is a pretty annoying footgun in go, but fortunately linters complain about it these days if missing with
> loopclosure: loop variable filter captured by func literal (govet)

neekolas: Go has some very unintuitive semantics around loops and closures. Without this, you'd end up always referencing the last fixture in the list. 

They are finally [fixing it](https://go.dev/wiki/LoopvarExperiment) but it's not widely enabled.

pkg/mls/api/v1/service.go
neekolas: Because Relay itself is loosely ordered it wouldn't change the overall guarantees. This particular case just leads to messages that might be _really_ out of order, instead of maybe off by one or two places.

snormore:
@neekolas @richardhuaaa Ready to merge this one?

richardhuaaa (approved):
Should be safe to merge!

PR #354, opened 2024-02-14
Add RPC requests limits & implement centralized subscription handler

tsachiherman:
## What ?

This PR addresses two separate issues:
1. It adds RPC handlers size limits.
2. Implement a single subscription handler for higher efficiency.

## RPC Requests Limits

This PR adds the following requests limits to the RPC handlers:
```golang
	// maxTopicsPerQueryRequest defines the maximum number of topics that can be queried in a single request.
	// the number is likely to be more than we want it to be, but would be a safe place to put it -
	// per Test_LargeQueryTesting, the request decoding already failing before it reaches th handler.
	maxTopicsPerQueryRequest = 157733

	// maxTopicsPerBatchQueryRequest defines the maximum number of topics that can be queried in a batch query. This
	// limit is imposed in additional to the per-query limit maxTopicsPerRequest.
	// as a starting value, we've using the same value as above, since the entire request would be tossed
	// away before this is reached.
	maxTopicsPerBatchQueryRequest = maxTopicsPerQueryRequest
```

On its own, it would "only" prevent DDoS attacks on the server; however, at scale, it would ensure the server could provide sufficient amount of resources per client.

Separately of this change, we need to update the monitoring system to detect the situations where we exceeded the thresholds and generate an alert.

## Subscription handler

The existing subscription model was quite naive; it would create a separate subscription for each topic. This was working well for small number of topics, but would completely fail with high number of topics.

To address that, I've created a subscription dispatcher, which is the *only* message subscriber. That handler compares each received message to any of the requested topics, and dispatch the message over a buffered channel.

### Before
```ssh
Benchmark_SubscribePublishQuery-10    	       1	31397530417 ns/op	87927504 B/op	 1346703 allocs/op
```

### After
```ssh
Benchmark_SubscribePublishQuery-10    	       1	1333448042 ns/op	60801280 B/op	 1014429 allocs/op
```
As a summary, you'd notice that there is a notable reduction in memory consumption ( 30% ).


tsachiherman:
@snormore could you please review the above briefly to see if I'm on the right track ? if so, I'll go ahead and create some unit tests for these changes.

pkg/api/message/v1/service.go
```diff
@@ -366,13 +431,35 @@ func (s *Service) BatchQuery(ctx context.Context, req *proto.BatchQueryRequest)
 		logFunc = log.Info
 	}
 	logFunc("large batch query", zap.Int("num_queries", len(req.Requests)))
-	// NOTE: in our implementation, we implicitly limit batch size to 50 requests
-	if len(req.Requests) > 50 {
-		return nil, status.Errorf(codes.InvalidArgument, "cannot exceed 50 requests in single batch")
+
+	// NOTE: in our implementation, we implicitly limit batch size to 50 requests (maxRequestLimitPerBatch = 50)
+	if len(req.Requests) > maxRequestLimitPerBatch {
```
snormore, line 436: Maybe call this `maxSubscribeRequestLimitPerBatch` so it's specific to subscribe

pkg/api/message/v1/service.go
```diff
@@ -35,6 +36,24 @@ const (
...
+
+	// maxRequestLimitPerBatch defines the maximum number of request we can support per batch.
+	maxRequestLimitPerBatch = 50
+
+	// maxTopicsPerRequest defines the maximum number of topics that can be queried in a single request.
+	maxTopicsPerRequest = 1024
+
+	// maxTopicsPerBatch defines the maximum number of topics that can be queried in a batch query. This
+	// limit is imposed in additional to the per-query limit maxTopicsPerRequest.
+	maxTopicsPerBatch = 4096
```
snormore, line 48: We have a [metric](https://app.datadoghq.com/dashboard/hmw-rv4-hxz/incident-15-investigation?fullscreen_end_ts=1707943281015&fullscreen_paused=false&fullscreen_refresh_mode=sliding&fullscreen_section=overview&fullscreen_start_ts=1707338481015&fullscreen_widget=8316307017714266&refresh_mode=sliding&view=spans&from_ts=1707856851255&to_ts=1707943251255&live=true) that shows up to 90k in the same request, so dropping it down to 4k would break that usage, and I doubt those clients have anything in place currently to auto-adjust down to something lower because it'd be a new error.

Curious what @neekolas thinks about having a limit here too.

pkg/api/message/v1/service.go
```diff
@@ -35,6 +36,24 @@ const (
...
+
+	// maxConcurrentSubscribedTopics is the total number of subscribed topics currently being registered by our node.
+	// request to surpass that limit would be rejected, as the server would not be able to scale to that limit.
+	maxConcurrentSubscribedTopics = 1024 * 1024
```
snormore, line 52: This depends on the available memory in the node, which changes depending on the deployment environment (devnet vs prod). Currently in prod we've seen over 6M concurrent topic subscriptions before memory alerts start firing.

Alternatively, or in additional to this, I think it would be great to limit concurrent topic subscriptions by IP. I think I'd prefer to start with that vs a global limit across all IPs, but I'm not against having a global limit either, it just might need to be configurable via CLI option instead of a const.

Curious about @neekolas's thoughts on this too.

snormore (commented):
Thanks for jumping on this @tsachiherman 🙏

pkg/api/message/v1/service.go
tsachiherman: could you tell me more about these clients ? are these the notification servers ?

pkg/api/message/v1/service.go
tsachiherman: Yes, I agree - we need multiple-tiers of limits to ensure the service won't be disrupted due to 1% of the clients placing unrealistic load on the server. I'm very much open to any other proposed default, which we would be able to dial-in later on. Does 100M sounds better as a starting point ?

tsachiherman: i.e. use a const for default, and override it via .env/config file/cli.

pkg/api/message/v1/service.go
snormore: The notification servers tend to use the SubscribeAll endpoint which just listens to the whole firehose rather than specific topics. These clients are more likely chatbots that listen to a lot of topics for their users. We could push them to use SubscribeAll but today they're listening to individual topics via this endpoint. I had a [wip branch](https://github.com/xmtp/xmtp-node-go/tree/snor/sub-nats-wildcard-for-many-topics) from a few weeks ago that updates this subscribe method to use the wildcard (listen to all topics) inside of this Subscribe endpoint to optimize memory usage (reduce open goroutines where there's a goroutine per topic subscription now), that could be worth implementing here instead of a limit on the batch size.

tsachiherman: ok, so I think that now I understand where the memory issue is coming from.

I was thinking about #topic ~= number of concurrent conversation in the inbox. but I wan't aware of chatbots.

These are completely different use case, with (potentially) different attributes : we might want to favor slower operation @ a lower resource consumption for that.

I'll spend some more time looking into that, but I do think your direction is the way to go.

pkg/api/message/v1/service.go
tsachiherman: I was thinking of much more fundamental change to address this level of scale. having go-routine per subscription is ridiculous and won't scale. Instead, I think that we should do the following:
1. create a single subscription per node, which would receive *all* the messages.
2. for each subscription (regardless of it's size), create a matching topics map.
3. when a message received, compare it to all the maps, after Unmarshal'ing it *once*, and dispatch it to all the recipients.

this approach is less efficient in quiescent scenarios, but would create a limit on the rendered dynamic system load.

pkg/api/message/v1/service.go
neekolas: Agree with @snormore that a hard limit isn't an option. If we were to change the behaviour for existing apps, we'd want to give them a very generous warning and a reasonable alternative.

pkg/api/message/v1/service.go
neekolas: It's going to be very difficult to determine what the right global limit is. Rejecting subscription requests might avoid one incident (nodes restarting), but it will create another incident (subscriptions mysteriously failing). 

Per IP, we can at least get a sense of what is reasonable and put a cap in so that 99th percentile user doesn't break the network for everyone.

neekolas:
With this model, do we actually need `nats` at all? It seems like the `subscriptionDispatcher` has a lot of overlap with what we were previously using `nats` for.

We would need support for MLS subscriptions to be integrated here as well, before we could outright shut nats down.

pkg/api/message/v1/service.go
tsachiherman: would you be ok if we'll start off with a very high number, and add configurable options later on ? i.e. start with a 1M limit ( far beyond what would crash the server )

pkg/api/message/v1/service.go
tsachiherman: the updated code doesn't reject any subscription requests anymore.

tsachiherman:
> With this model, do we actually need `nats` at all? It seems like the `subscriptionDispatcher` has a lot of overlap with what we were previously using `nats` for.
> 
> We would need support for MLS subscriptions to be integrated here as well, before we could outright shut nats down.

I think that we'll still need it as long as we have multiple node instances, since it's the message synchronization layer, right ?

Even with MLS, we might want to keep it ( or its functional equivalent ) to support horizontal scaling. ( am I missing something ? )

pkg/api/message/v1/service.go
tsachiherman: The above metric mention by @snormore applies to the subscribe, not for the query, btw. We don't have a metric, afaik, that tracks the query size.

neekolas:
> I think that we'll still need it as long as we have multiple node instances, since it's the message synchronization layer, right ?

The way we use `nats` here is just as a local subscription dispatcher. Message synchronization between nodes happens via Waku Relay (which we are going to need to keep)

pkg/api/message/v1/service.go
tsachiherman: I updated the numerics here per a unit-test created, where I tested the max possible topics count.
This would guarantee it's won't break anything for the time being, and we will be able to adjust it in the future.

pkg/api/message/v1/service.go
```diff
@@ -366,13 +377,35 @@ func (s *Service) BatchQuery(ctx context.Context, req *proto.BatchQueryRequest)
 		logFunc = log.Info
 	}
 	logFunc("large batch query", zap.Int("num_queries", len(req.Requests)))
-	// NOTE: in our implementation, we implicitly limit batch size to 50 requests
-	if len(req.Requests) > 50 {
-		return nil, status.Errorf(codes.InvalidArgument, "cannot exceed 50 requests in single batch")
+
+	// NOTE: in our implementation, we implicitly limit batch size to 50 requests (maxSubscribeRequestLimitPerBatch = 50)
+	if len(req.Requests) > maxSubscribeRequestLimitPerBatch {
```
neekolas, line 382: `BatchQuery` and subscribe are unrelated endpoints. I'd suggest we give each their own constants/limits.

pkg/api/message/v1/service.go
neekolas: Something like 1M per request would be fine. It's way above what any valid user needs.

If we were to enforce the limit per IP, we'd just have to be very sure the bookkeeping is correct. For example, that a client can't subscribe to 100,000 topics, disconnect, and resubscribe to the same list and push themselves over the limit.

pkg/api/message/v1/service.go
tsachiherman: I believe that @snormore asked for this variable name, although I'm very much open to renames. Do you want to suggest a different name ?

neekolas: Maybe `maxQueriesPerBatch`

pkg/api/message/v1/service.go
tsachiherman: Agree - IP filtering isn't trivial. But it's also a bit beyond the scope of this PR. The goal here was to apply some basic and immediate remedies to the existing (crashing) node on the "happy" path.
I can look into IP filtering next.

pkg/api/message/v1/service.go
tsachiherman: ok, done.

tsachiherman:
> > I think that we'll still need it as long as we have multiple node instances, since it's the message synchronization layer, right ?
> 
> The way we use `nats` here is just as a local subscription dispatcher. Message synchronization between nodes happens via Waku Relay (which we are going to need to keep)

Oh.. yes. I wouldn't have found that if not being told.. it's quite a spaghetti path.
Mind if we'll do that on a separate PR ? this one seems to be big enough already, and I want to avoid monster PRs.

neekolas:
Given the scope of the changes here, I'd suggest we push this up as a docker image and try running one of our client SDK's test suites against it (maybe `xmtp-js`). That should validate that things are at least roughly working as expected.

neekolas:
> Mind if we'll do that on a separate PR ? this one seems to be big enough already, and I want to avoid monster PRs.

Absolutely. That one is going to be delicate, since we need to make sure that messages still make it from Waku Relay -> apiV1 AND the MLS API, which `nats` is currently facilitating.

tsachiherman:
> Given the scope of the changes here, I'd suggest we push this up as a docker image and try running one of our client SDK's test suites against it (maybe `xmtp-js`). That should validate that things are at least roughly working as expected.

That makes sense, and I'd like to pursue that.. but not sure how to. Is there any pipeline/github action that does that ?

snormore:
The lint error in the CI pipeline for this PR is completely unrelated to your changes. I'm not sure how it made it's way into a green `main` tbh, but you can fix with https://github.com/xmtp/xmtp-node-go/pull/351/commits/cf16a9528604664caa119ecd125b04c83eef5834 (which I pushed to [one of my open PRs](https://github.com/xmtp/xmtp-node-go/pull/351) too).

snormore:
You can repurpose [this github workflow](https://github.com/xmtp/xmtp-node-go/blob/main/.github/workflows/push-mls.yml) if you want to build+push the image with another tag from this branch.

pkg/api/message/v1/service.go
```diff
@@ -35,6 +34,20 @@ const (
...
 	// 1048576 - 300 - 62 = 1048214
 	MaxMessageSize = pubsub.DefaultMaxMessageSize - MaxContentTopicNameSize - 62
+
+	// maxQueriesPerBatch defines the maximum number of queries we can support per batch.
+	maxQueriesPerBatch = 50
+
+	// maxTopicsPerRequest defines the maximum number of topics that can be queried in a single request.
+	// the number is likely to be more than we want it to be, but would be a safe place to put it -
+	// per Test_LargeQueryTesting, the request decoding already failing before it reaches th handler.
+	maxTopicsPerRequest = 157733
```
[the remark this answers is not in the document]
tsachiherman: But.. it's not per batch. it's used both in the `Query` endpoint as well as in the `BatchQuery`.
The subsequent one, `maxTopicsPerBatch` is the one that is being used per batch.

pkg/api/message/v1/service.go
```diff
@@ -366,13 +381,35 @@ func (s *Service) BatchQuery(ctx context.Context, req *proto.BatchQueryRequest)
...
+	for _, query := range req.Requests {
+		totalRequestedTopicsCount += len(query.ContentTopics)
+	}
+
+	if totalRequestedTopicsCount == 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "content topics required")
+	}
+
+	// are we still within limits ?
+	if totalRequestedTopicsCount > maxTopicsPerBatch {
```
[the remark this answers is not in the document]
tsachiherman: maybe the variable name `maxTopicsPerBatch` wasn't clear enough : it should be the limit for the total number of topics per batch query. I'll update the variable name accordingly.

pkg/api/message/v1/service.go
```diff
@@ -366,13 +381,35 @@ func (s *Service) BatchQuery(ctx context.Context, req *proto.BatchQueryRequest)
...
+		log.Info("batch query exceeded topics count threshold", zap.Int("topics_count", totalRequestedTopicsCount))
```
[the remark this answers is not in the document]
tsachiherman: Yes, I'm ok dropping this one, but I'd like to modify the reported error to include the topics count then.

pkg/api/message/v1/service.go
```diff
@@ -366,13 +381,35 @@ func (s *Service) BatchQuery(ctx context.Context, req *proto.BatchQueryRequest)
...
+		return nil, status.Errorf(codes.InvalidArgument, "batch content topics count exceeded maximum topics per batch threshold of %d", maxTopicsPerBatch)
 	}
+
 	// Naive implementation, perform all sub query requests sequentially
 	responses := make([]*proto.QueryResponse, 0)
 	for _, query := range req.Requests {
+		if len(query.ContentTopics) > maxTopicsPerRequest {
```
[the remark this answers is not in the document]
tsachiherman: I think not.. it should align with the same limits we have in `Service.Query`. ( where we had a single query ).

pkg/api/message/v1/subscription.go
```diff
@@ -0,0 +1,156 @@
...
+			case subscription.messagesCh <- &env:
+			default:
+				// we got here since the message channel was full. This happens when the client cannot
+				// consume the data fast enough. In that case, we don't want to block further since it migth
+				// slow down other users. Instead, we're going to close the channel and let the
+				// consumer re-establish the connection if needed.
+				close(subscription.messagesCh)
+				delete(d.subscriptions, subscription)
+			}
+			continue
```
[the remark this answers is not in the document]
tsachiherman: good find!

pkg/api/message/v1/service.go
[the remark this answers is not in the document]
snormore: Ok I think maybe I'm interpreting the terms differently; you're saying that the grpc request is the batch and that there are multiple "requests" within that. That seems reasonable.

pkg/api/message/v1/service.go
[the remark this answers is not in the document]
snormore: I see, the double use of the term request (as in grpc request and query request) was confusing me. Maybe we can prefix these request references with query to make it clearer, like `QueryRequest`?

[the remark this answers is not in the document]
tsachiherman: I already modified it to `maxTopicsPerQueryRequest` and `maxTopicsPerBatchQueryRequest`. do you think it's clear enough ?

snormore (approved):

Issue #342, opened 2024-01-23
Feature request: Metrics for MLS interface


snormore:
### Is your feature request related to a problem?

The MLS API was added but does not yet include any telemetry/metrics.

### Describe the solution to the problem

We should emit metrics similar to the messagev1 API, for
- API requests/errors and duration, including app/client version labels
- Subscriptions
- Sent messages

PR #355, opened 2024-02-20
Emit metrics for MLS sent messages

snormore:
Addresses part of https://github.com/xmtp/xmtp-node-go/issues/342


tsachiherman (approved):
LGTM

neekolas (approved):
Nice! Thanks for adding this. We're definitely going to need to keep adding observability during the alpha.

New MLS E2E tests is somewhere on my agenda as well.

PR #361, opened 2024-02-27
Restore centralized subscription handler

richardhuaaa:
1. Restore the centralized subscription handler (#354)
2. Remove the RPC request limits that were added - to reduce number of variables (can add separately)
3. Include @neekolas 's additional logging
4. Log lots of info when a subscription channel is full, but don't close it
5. 4x the backlog length for subscribeAll

### DIFF
--- a/pkg/api/message/v1/subscription.go
+++ b/pkg/api/message/v1/subscription.go
@@ -13,7 +13,7 @@ import (
 
 const (
 	// allTopicsBacklogLength defines the buffer size for subscriptions that listen to all topics.
-	allTopicsBacklogLength = 1024
+	allTopicsBacklogLength = 4096
 
 	// minBacklogBufferLength defines the minimal length used for backlog buffer.
 	minBacklogBufferLength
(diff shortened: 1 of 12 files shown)


pkg/api/message/v1/subscription.go
```diff
@@ -0,0 +1,157 @@
...
+			break
+		}
+	}
+	if !sub.all {
+		sub.topics = topics
+		// use a log2(length) as a backbuffer
+		backlogBufferSize := log2(uint(len(topics))) + 1
+		if backlogBufferSize < minBacklogBufferLength {
+			backlogBufferSize = minBacklogBufferLength
+		}
```
richardhuaaa, line 136: I'm wondering if this was the source of the issue. `minBacklogBufferLength` doesn't seem to have been set

neekolas: In go constants get automatically set to the value in the constant above. In this case, the subscribe all limit

richardhuaaa: I see! Okay I set it explicitly back to 1024

neekolas (approved):

PR #366, opened 2024-03-28
Use a shared nats server

neekolas:
## Summary

There is an issue with `SubscribeAll` and group messages. Both the legacy and the MLS APIs create their own nats servers. That means that the servers cannot talk to one another. This only becomes an issue with `SubscribeAll`, which is expected to receive messages from both servers.

- This PR creates a shared nats server for both APIs and passes it in as an argument

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -12,6 +12,7 @@ import (
 	middleware "github.com/grpc-ecosystem/go-grpc-middleware"
 	prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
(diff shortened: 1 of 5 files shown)


nplasterer (approved):

pkg/api/server.go
```diff
@@ -253,6 +267,9 @@ func (s *Server) startHTTP() error {
 
 	tracing.GoPanicWrap(s.ctx, &s.wg, "http", func(ctx context.Context) {
 		s.Log.Info("serving http", zap.String("address", s.httpListener.Addr().String()))
+		if s.httpListener == nil {
```
neekolas, line 270: The race test was catching on this. Surprised that wasn't causing failures earlier.

PR #371, opened 2024-04-11
Scaffold identity server

richardhuaaa:
This generates the protos, and stubs out the identity service API endpoints. I'll work on the DB changes next.

The existing unit tests pass, but to verify this actually did something, is the idea to land this, and then run curl against the endpoints on the GRPC gateway on dev?


dev/up
```diff
@@ -8,7 +8,7 @@ if ! which golangci-lint &>/dev/null; then brew install golangci-lint; fi
 if ! which shellcheck &>/dev/null; then brew install shellcheck; fi
 if ! which protoc &>/dev/null; then brew install protobuf; fi
 if ! which protoc-gen-go &>/dev/null; then go install google.golang.org/protobuf/cmd/protoc-gen-go@latest; fi
-if ! which mockgen &>/dev/null; then go install go.uber.org/mock/mockgen@latest; fi
```
richardhuaaa, line 11: This one was a pain to find, but tests were failing until I did this.

Basically, it's entirely possible you have a different version of `mockgen` installed on your system (in fact, you're guaranteed to have it if you have developed against `xmtp-node-go` in the past). That would be github.com/golang/mock/mockgen instead of this uber one.

The reason for the use of the uber one seems to be the `.Satisfied` property, which lets you check on an interval if all of the expected API calls were performed on the mock. We're using it in tests like this:

```
require.Eventually(t, ctrl.Satisfied, 5*time.Second, 100*time.Millisecond)
```

neekolas: Nice catch

dev/up
```diff
@@ -8,7 +8,7 @@ if ! which golangci-lint &>/dev/null; then brew install golangci-lint; fi
...
+go install go.uber.org/mock/mockgen@latest
```
neekolas, line 11: Is there no way to tell if you have the correct mockgen and only run this if you have the wrong one?

README.md
```diff
@@ -15,6 +15,8 @@ After `xmtpd` meets specific functional requirements, the plan is for it to beco
 - [Go](https://go.dev/doc/install)
 - [Docker](https://www.docker.com/get-started/)
 
+You must have the _exact_ go version listed in `go.mod` - you can verify this by running `go version`.
```
neekolas, line 18: Of all our projects, this is probably the one that needs a devcontainer the most come to think of it.

neekolas (approved):

dev/up
richardhuaaa: Okay, I found a hack to only run this if the version is not 0.4.0. Basically we can pin the uber mockgen on the current version of v0.4.0 and update as needed. Should be good indefinitely because it is running behind the golang one, which is on v1.6.0

neekolas: Nice work!

PR #373, opened 2024-04-11
Allow larger queries for user preferences

neekolas:
## tl;dr

We are putting in this hack to allow for larger request sizes for User Preferences requests.

This will be removed once MLS/Groups moves into production and we can start caching these results locally.

### DIFF
--- a/pkg/api/message/v1/service.go
+++ b/pkg/api/message/v1/service.go
@@ -326,12 +326,13 @@ func (s *Service) SubscribeAll(req *proto.SubscribeAllRequest, stream proto.Mess
 func (s *Service) Query(ctx context.Context, req *proto.QueryRequest) (*proto.QueryResponse, error) {
 	log := s.log.Named("query").With(zap.Strings("content_topics", req.ContentTopics))
 	log.Debug("received request")
+	numContentTopics := len(req.ContentTopics)
 
-	if len(req.ContentTopics) == 0 {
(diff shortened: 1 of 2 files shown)


pkg/api/message/v1/service.go
```diff
@@ -395,3 +399,11 @@ func (s *Service) BatchQuery(ctx context.Context, req *proto.BatchQueryRequest)
 		Responses: responses,
 	}, nil
 }
+
+func getMaxRows(contentTopics []string) int {
+	if len(contentTopics) == 1 && topic.IsUserPreferences(contentTopics[0]) {
```
nplasterer, line 404: Maybe put a Todo or comment in here to mention that this is temporary and should be removed. Sometimes I like adding a link to the PR as well for future people/me.

nplasterer (approved):

nplasterer:
Just adding some additional color to the issue this is solving. In the absence of persistence across client creations in our mobile SDKs the consent records along with the conversation records must be loaded in it's entirety every time the app is opened. This is incredibly non performant. But saw great performance improvements by growing the page size. Persistence for conversations and their consent records will be added as part of the MLS 1to1 messages and this should no longer be needed.

alexrisch (approved):

pkg/api/message/v1/service.go
neekolas: Good call

PR #374, opened 2024-04-12
Identity database schema and publishing algorithm

richardhuaaa:
I would love feedback on the publishing algorithm and resulting schema, if I've missed an edge case or if anything has been over-complicated. I will polish and publish the description somewhere more logical in the next PR.

I've optimized for:
- More expensive writes in exchange for more performant reads
- Support for high concurrency/reentrancy
- Consistency and correctness

Being able to resume a publish after the server goes down is a side effect and not a primary goal


richardhuaaa:
The other thing to consider is with any event log like this, validation on each subsequent publish gets more expensive as the log gets longer. Is there a denial-of-service attack possible here by repeatedly appending to one inbox_id?

neekolas:
> The other thing to consider is with any event log like this, validation on each subsequent publish gets more expensive as the log gets longer. Is there a denial-of-service attack possible here by repeatedly appending to one inbox_id?

Completely agree. We are going to have to have some sort of overall limit to the length of the event log. Not just a denial of service on the nodes, but also on other group members who have to download and process these logs.

50? 100?

pkg/migrations/mls/20240411200242_init-identity.up.sql
```diff
@@ -0,0 +1,34 @@
...
+    identity_update_proto BYTEA NOT NULL
+);
+
+--bun:split
+
+CREATE INDEX idx_inbox_log_inbox_id_commit_status ON inbox_log(inbox_id, commit_status);
+
+--bun:split
+
+CREATE TABLE address_log (
```
neekolas, line 19: I'm wondering where the address_log actually gets queried. It seems like all our APIs use either the `inbox_log` or the `address_lookup_cache`

pkg/migrations/mls/20240411200242_init-identity.up.sql
```diff
@@ -0,0 +1,34 @@
...
+    inbox_log_sequence_id BIGINT PRIMARY KEY,
```
neekolas, line 20: In the notes above it says 

> the address_lookup_cache stores the inbox_log_sequence_id

But you have that field only on the `address_log` here.

pkg/identity/api/v1/service.go
```diff
@@ -41,14 +41,45 @@ func (s *Service) Close() {
 	s.log.Info("closed")
 }
 
+/*
+Algorithm:
+
+Note - the inbox_log table has global ordering via a serial sequence_ID, and an index by inbox_ID
```
neekolas, line 47: I think we might be able to get to the same outcome more simply if we just lean on Postgres' serializable transactions. Would look something like this pseudo-sql

```sql
BEGIN
SELECT * FROM inbox_log FOR UPDATE WHERE inbox_id = ?;
-- Take the selected rows and send them to the MLS validation service to validate and get the results
-- If invalid, abort the transaction and return an error

-- For each updated address, update the address_lookup_cache and address_log
INSERT INTO address_log VALUES(...)
UPDATE address_lookup_cache SET inbox_id = ? WHERE address = ?
COMMIT
```

If another update is attempting to write to the address lookup cache, you should hit a serialization error and be forced to retry.

neekolas: You can read more about the guarantees of serializable transactions [here](https://www.thenile.dev/blog/transaction-isolation-postgres)

pkg/migrations/mls/20240411200242_init-identity.up.sql
```diff
@@ -0,0 +1,34 @@
...
+    identity_update_proto BYTEA NOT NULL
+);
+
+--bun:split
+
+CREATE INDEX idx_address_log_address ON address_log(address);
+
+--bun:split
+
+CREATE TABLE address_lookup_cache (
```
neekolas, line 31: I wonder if we can merge the `address_lookup_cache` and the `address_log` into one thing.

What if we had a table like this:

```sql
CREATE TABLE address_history (
     address TEXT NOT NULL,
     inbox_id TEXT NOT NULL,
     inbox_log_sequence_id BIGINT NOT NULL,
     
     PRIMARY KEY (address, inbox_log_sequence_id)
)
```

This would be an append-only table, where we would write new entries on each valid update that touches an address. It gives us a history of updates, and makes it easy to query for the current update

To read the current value we would do something like 
```sql
SELECT DISTINCT ON (address) inbox_id WHERE address IN (?) ORDER BY inbox_log_sequence_id DESC;
```

With the right index on address/inbox_log_sequence_id it should still be very fast to read.

neekolas: Or maybe a GROUP BY, which is usually faster. But the same idea

neekolas: We'd have to benchmark and see if it makes the wrong trade-off in terms of read vs write performance. But if it's using the indexes correctly, it might be roughly equivalent.

pkg/identity/api/v1/service.go
richardhuaaa: My hang-up with transactions was that I was uncertain if the use of a global SERIAL ID would cause all transactions to block on each other whenever a row was inserted. From our discussion, we found that sequences increment outside of transactions, and if a transaction rolls back, the sequence can have gaps - which is exactly the type of behavior we want.

pkg/identity/api/v1/service.go
```diff
@@ -41,14 +41,45 @@ func (s *Service) Close() {
...
+ 2. Read the log for the inbox_id, ordering by sequence_id
+ 3. If the log has more than 256 entries, abort the transaction.
+ 3. Validate it sequentially. If failed, abort the transaction.
+ 4. For each affected address:
+    a. Insert the update into the address_log table
+    -- Note: There may be races across multiple inboxes. The address_log can use
+    -- inbox_log_sequence_ID to establish ordering.
+    b. Read the log for the address, ordering by *inbox_log_sequence_id*
+    c. If the log has more than 256 entries, abort the transaction.
+    c. Process the address log (without signature validation) and compute the final inbox ID
```
neekolas, line 59: There's really only two things we need to check here:
- That there are < 256 updates in the log
- That the log is either empty or the last message is a revocation

We might be able to get away with just a `COUNT` query and pulling the last record, and skip all the middle entries.

pkg/migrations/mls/20240411200242_init-identity.up.sql
```diff
@@ -0,0 +1,28 @@
+SET statement_timeout = 0;
+
+--bun:split
+
+CREATE TABLE inbox_log (
+    sequence_id BIGSERIAL PRIMARY KEY,
+    inbox_id TEXT NOT NULL,
```
neekolas, line 7: One performance optimization we _could_ do here:

We could shrink down some indexes if we had an `inbox` table, and just used a foreign key. `inbox_id`s are 64 bytes and a BIGSERIAL field is 8.

pkg/identity/api/v1/service.go
richardhuaaa: I'm thinking about this scenario:

```
Associate address to inbox 1
Associate address to inbox 2
Revoke address from inbox 1
```

richardhuaaa: It's not so much verifying, as figuring out what the correct inbox ID should be based on the entries

pkg/migrations/mls/20240411200242_init-identity.up.sql
richardhuaaa: Would the `inbox` table just have `id` and `inbox_id` columns?

I'm not sure that would shrink the size - indexes store a hash of the column, not the raw column, right?

richardhuaaa: All queries will need to resolve the inbox log entry starting from the `inbox_id` - so I think that just moves the same index onto a different table, and writes will need to update both tables regardless? I guess we could consider using a [HASH index](https://www.postgresql.org/docs/current/hash-intro.html#:~:text=Hash%20indexes%20support%20only%20single,not%20the%20actual%20column%20value.) instead of a BTREE index, assuming we have the [right postgres version](https://hakibenita.com/postgresql-hash-index#hash-index-in-postgresql)

richardhuaaa: I guess one thing to be careful about - the first link says `As a result of the overflow cases, we can say that hash indexes are most suitable for unique, nearly unique data or data with a low number of rows per hash bucket`. I suppose most inbox_id's will have a low number of rows, but I'm unsure what the limit of 'low' is

pkg/identity/api/v1/service.go
neekolas: Ohhh, I see. What I proposed only works for adds but doesn't help at all for revocation

pkg/migrations/mls/20240411200242_init-identity.up.sql
neekolas: > so I think that just moves the same index onto a different table, and writes will need to update both tables regardless?

The main advantage is that you only need to store/index it once, instead of storing/indexing the full size per update

pkg/identity/api/v1/service.go
neekolas: Thinking about the revocation case a bit more.

What about this model:

```sql
CREATE TABLE address_log (
    associated_at_inbox_log_sequence_id BIGINT NOT NULL,
    revoked_at_inbox_log_sequence_id BIGINT,
    address TEXT NOT NULL,
    inbox_id TEXT,
);
```

When a revocation happens we just update the previous association by setting the `revoked_at_inbox_log_sequence_id` field.

We can then query it by only looking for results where the `revoked_at_inbox_log_sequence_id` field is null and get the last still-valid association.

neekolas: Just avoids the need to re-process the address log on updates, which seems complicated since the individual entries only make sense in the context of an `identity_log` history

neekolas: I don't love that this is still not purely append-only, which would be the easiest to decentralize.

pkg/api/message/v1/service.go
```diff
@@ -324,9 +324,12 @@ func (s *Service) SubscribeAll(req *proto.SubscribeAllRequest, stream proto.Mess
 }
 
 func (s *Service) Query(ctx context.Context, req *proto.QueryRequest) (*proto.QueryResponse, error) {
-	log := s.log.Named("query").With(zap.Strings("content_topics", req.ContentTopics))
-	log.Debug("received request")
+	log := s.log.Named("query")
 	numContentTopics := len(req.ContentTopics)
+	if numContentTopics < 200 {
```
neekolas, line 329: I like this change, but did you mean to put it in this PR

pkg/identity/api/v1/service.go
richardhuaaa: I like this, it's more scalable. I generalized it a little to handle cases where there are multiple associations and revocations

pkg/migrations/mls/20240411200242_init-identity.up.sql
richardhuaaa: Is this a common pattern? I'm not sure whether that would be more performant than what postgres does internally with indices (btree or hash, depending on what we want to do). Are we concerned about size from the perspective of latency, or is it intrinsically about size? For latency, note that having an extra table would require an extra JOIN on every read and write.

pkg/api/message/v1/service.go
richardhuaaa: Yes, I figured it was small enough that it didn't matter too much. The tests were failing with unreadable logs before this change

pkg/migrations/mls/20240411200242_init-identity.up.sql
richardhuaaa: I guess another way of looking at it - do we prefer our data normalized, or denormalized, with the classic tradeoffs involved? A separate table would be more normalized, having it built into one table is more denormalized

neekolas: If we had an old-school DBA here they'd say we should normalize everything we can and that's how SQL databases are designed to work. 

These days the databases are so powerful that a lot of apps optimize for simplicity and denormalize more...and it's usually OK. There's a performance cost, but not enough to cause problems until you get to a really large scale.

neekolas (approved):

pkg/migrations/mls/20240411200242_init-identity.up.sql
neekolas: Summarizing an offline conversation:

We've decided to leave the schema as-is. We'll trade larger index sizes for simpler read queries, which makes sense given that the data isn't duplicated that often (most users will have relatively few identity updates)

pkg/identity/api/v1/service.go
richardhuaaa: Coming back to this days later, this causes a different problem - the sequence number is not guaranteed to increase in the same 'serialization order' as the transactions: https://stackoverflow.com/questions/52432459/postgresql-serialized-inserts-interleaving-sequence-numbers

Thinking through how we can fix this and will put up another PR

PR #376, opened 2024-04-19
Remove first two unused indexes

neekolas:
## tl;dr

We have a bunch of gigantic Postgres indexes that are completely unused. I am dropping the first three that feel uncontroversial.

## Some data

These are our current database indexes

By running this SQL, I got the list of when they were last read:

```sql
SELECT relname , indexrelname , idx_scan , idx_tup_read , idx_tup_fetch
FROM pg_stat_user_indexes
WHERE schemaname = 'public' and relname = 'message';
```

Which tells me that this index has _never_ been used in a query.

`messageindex` is used as a primary key so we can't kill it.

## Next up

There are three more indexes that have been read in the past, but when I re-run the query above the values don't change or change very little.

`message_recvts_shouldexpire_idx` gets used extremely infrequently, as it has largely been replaced with `message_recvts_shouldexpiretrue_idx` which gets used a lot. But I'd like to understand why the delta isn't 0.

`message_contenttopic_idx` is also quite infrequently read. Because it shares a prefix with `message_ctopic_sts_id_idx`, I believe deleting it is safe and the other index could be used. But I'm also hesitant there.

`message_sort_idx` has basically never been used and is safe to delete.


neekolas:
* **#376** <a href="https://app.graphite.dev/github/pr/xmtp/xmtp-node-go/376?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pkg/migrations/messages/20240419172419_remove-unused-indexes-pt-1.down.sql
```diff
@@ -0,0 +1,5 @@
+SET statement_timeout = 0;
+
+--bun:split
+
+CREATE INDEX CONCURRENTLY IF NOT EXISTS message_pubsubTopic_idx ON public.message (pubsubTopic);
```
richardhuaaa, line 5: Without CONCURRENTLY, does the server start up extremely slowly?

pkg/migrations/messages/20240419172419_remove-unused-indexes-pt-1.up.sql
```diff
@@ -0,0 +1,4 @@
+SET statement_timeout = 0;
+
+DROP INDEX CONCURRENTLY IF EXISTS messageindex;
```
richardhuaaa, line 3: I thought the description said this one can't be killed because it's a primary key?

richardhuaaa (approved):

pkg/migrations/messages/20240419172419_remove-unused-indexes-pt-1.down.sql
neekolas: It would also take a lock of the entire table and block reads

neekolas: I'm having problems actually running concurrent index drops in a migration (for some reason it is trying to run the migration in a tx, which is not allowed). I might just make this non-concurrent but manually drop the index concurrently in prod first.

neekolas:
I've manually dropped `message_pubsubTopic_idx` and `message_sort_idx` concurrently in prod. Will do the same in dev and then update this PR to include both drops.

PR #377, opened 2024-04-21
Implement PublishIdentityUpdate and GetInboxLogs endpoints

richardhuaaa:
Stubbed out the validation service for now, as well as the `GetInboxIds` endpoint - these will be implemented in the next PR. Also will add more efficient InboxID encoding, and idempotent publishes, shortly.

Notes:

1. I did not split the logic between 'store' and 'service' layers. It seems 90% of the logic would be in the database anyway, and it felt like splitting it up would make it harder to optimize later. Happy to refactor if anyone disagrees though.

2. Our ORM Bun does not currently support [custom isolation levels](https://github.com/uptrace/bun/blob/b005dc2a9034715c6f59dcfc8e76aa3b85df38ab/driver/pgdriver/driver.go#L218) for Postgres. Decided to punt this for now - using SERIALIZABLE isolation level just prevents races when two `CreateInbox` updates are published simultaneously, which should be mostly prevented by our hash validation mechanism anyway.


pkg/mls/store/store.go
```diff
@@ -53,6 +64,111 @@ func New(ctx context.Context, config Config) (*Store, error) {
...
 }
 
+func (s *Store) PublishIdentityUpdate(ctx context.Context, req *identity.PublishIdentityUpdateRequest) (*identity.PublishIdentityUpdateResponse, error) {
+	new_update := req.GetIdentityUpdate()
+	if new_update == nil {
+		return nil, errors.New("IdentityUpdate is required")
+	}
+
+	// TODO: Implement serializable isolation level once supported
+	if err := s.db.RunInTx(ctx, &sql.TxOptions{ /*Isolation: sql.LevelSerializable*/ }, func(ctx context.Context, tx bun.Tx) error {
```
neekolas, line 74: Why is every tool for working with SQL in Go so terrible?

pkg/mls/store/store.go
```diff
@@ -53,6 +64,111 @@ func New(ctx context.Context, config Config) (*Store, error) {
...
+}
+
+func (s *Store) GetInboxLogs(ctx context.Context, batched_req *identity.GetIdentityUpdatesRequest) (*identity.GetIdentityUpdatesResponse, error) {
+	reqs := batched_req.GetRequests()
+	resps := make([]*identity.GetIdentityUpdatesResponse_Response, len(reqs))
+
+	for i, req := range reqs {
+		inbox_log_entries := make([]*InboxLogEntry, 0)
+
+		err := s.db.NewSelect().
```
neekolas, line 138: Would be nice to get all of these in a single SQL query, but we can optimize that later.

neekolas (approved):

pkg/mls/store/store.go
neekolas: I hate Bun. I hate Gorm. Sqlx is fine but very bare-bones.

I've heard good things about [`sqlc` but never tried it](https://sqlc.dev/)

richardhuaaa: Awesome, this will be a good list to try when we come back to this. Always hate working with ORM's

PR #378, opened 2024-04-23
Add support for validation service

richardhuaaa:
This just implements the wrappers for the validation service, and plumbs it into the identity service. I will be out for the next two weeks, feel free to modify or land on my behalf.

I wasn't able to complete the remaining piece of the identity service, this is what needs to be done in the next PR:

1. Implement the TODO's [here](https://github.com/xmtp/xmtp-node-go/blob/9cb29579f8225932d7ce25bc80697aed3743b7d7/pkg/mls/store/store.go#L100) and [here](https://github.com/xmtp/xmtp-node-go/blob/9cb29579f8225932d7ce25bc80697aed3743b7d7/pkg/mls/store/store.go#L121), using the logic described [here](https://github.com/xmtp/xmtp-node-go/blob/9cb29579f8225932d7ce25bc80697aed3743b7d7/pkg/identity/api/v1/identity_service.go#L70-L76). The validation can be done by calling into the handler introduced in this PR.
2. Implement this [read](https://github.com/xmtp/xmtp-node-go/blob/9cb29579f8225932d7ce25bc80697aed3743b7d7/pkg/identity/api/v1/identity_service.go#L93)


neekolas (approved):
This looks safe to build on top of. Think we can merge and keep going to actually integrate the validation service into `PublishIdentityUpdate
`

37ng (approved):

insipx (approved):

PR #379, opened 2024-04-24
GetInboxIds

insipx:
- still working through some tests/sql


pkg/mls/store/store.go
```diff
@@ -169,6 +170,54 @@ func (s *Store) GetInboxLogs(ctx context.Context, batched_req *identity.GetIdent
...
+	db_log_entries := make([]*GroupedAddressLogEntry, 0)
+
+	// maybe restrict select to only fields in groupby?
+	err := s.db.NewSelect().
+		Model((*AddressLogEntry)(nil)).
+		ColumnExpr("address, inbox_id, MAX(association_sequence_id) as max_association_sequence_id").
+		Where("address IN (?)", bun.In(addresses)).
+		Where("revocation_sequence_id IS NULL OR revocation_sequence_id < association_sequence_id").
+		Group("address", "inbox_id").
+		OrderExpr("MAX(association_sequence_id) ASC").
```
neekolas, line 195: Right, this is what we were missing this morning

insipx: The query is a bit hacky. In order to get the right inbox id (with the highest association_sequence_id) I changeed the order to "ASC", which the second-order inner for-loop subsequently chooses to be the right inboxId since it overwrites the previous entry.

Could use some better thinking there

neekolas: I think you can make it work all in SQL by joining the table to itself.

Something like:

```sql
SELECT 
    a.address,
    a.inbox_id,
    a.association_sequence_id
FROM 
    address_log a
INNER JOIN (
    SELECT 
        address,
        MAX(association_sequence_id) AS max_association_sequence_id
    FROM 
        address_log
    WHERE 
        address = ANY(?)
    GROUP BY 
        address
) b ON a.address = b.address AND a.association_sequence_id = b.max_association_sequence_id;
```

insipx:
close in favor of #382 & #383

PR #380, opened 2024-04-25
Migrate MLS DB To SQLC

neekolas:
## tl;dr

Bun has been bothering me for a while, so I've been working on a migration that will get rid of our ORM altogether and just use boring SQL queries for everything.

[`sqlc`](https://sqlc.dev/) is a very slick tool to generate code based on plain SQL queries using placeholders for arguments. It's not perfect...I had to do some gymnastics to make a few of the query types work. 

But the fact that there is no runtime other than the standard SQL driver and some generated code outweighs its limitations IMO. There's no fancy ORM library to worry about mangling your queries, and the learning curve is basically just "how well do you know SQL".

## What's wrong with Bun?
- No support for serializable transactions
- The SQL driver is not as well maintained as PGX
- High learning curve to build complex queries, even if you know SQL well
- Relations system is not very powerful and ends up doing N+1 queries a lot of the time.
- Configuring the database with struct tags is errorprone, and there aren't great checks to make sure the struct tags actually match the schema.

## Things that suck right now with sqlc

- I can't find a good way to have dynamic ORDER BY expressions. So I literally have separate queries for ASC and DESC versions. It's not the end of the world, but it's very frustrating. There's an [issue to fix it](https://github.com/sqlc-dev/sqlc/issues/2061), and some hacky workarounds using CASE statements, but it's not great.
- Making the filters play nice with `json_populate_recordset` is a bit of a pain. Switching to the `pgx` driver helped, since I think there was a bug in Bun's pgdriver.

## Migration plan
We use Bun in a lot of places and for a lot of things today.

- It powers the `authz` database and all the migrations there
- It powers the migrations for the `message` database (but not the queries)
- It powers the `mls` database and all the queries in the `mlsstore`.

The priority right now is to remove it from the `mlsstore`. We will still use it for migrations (`sqlc` can read Bun migrations just fine).

This involves replacing the bun `pgdriver` with `pgx` (done in this PR) and replacing all the Bun ORM queries with `sqlc` queries. I have most of the queries written, but I'll split up the actual migration over several PRs. This can be done incrementally, but once the process is complete we can delete the Bun models.

We aren't using any of the fancy `sqlc` cloud features and have no plans to. 

## What knucklehead brought Bun into our codebase?

Ummmm. 😬. That was me.


neekolas:
* **#381** <a href="https://app.graphite.dev/github/pr/xmtp/xmtp-node-go/381?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#380** <a href="https://app.graphite.dev/github/pr/xmtp/xmtp-node-go/380?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pkg/mls/store/queries.sql
```diff
@@ -0,0 +1,78 @@
+-- name: GetAllInboxLogs :many
+SELECT * FROM inbox_log
+WHERE inbox_id = $1
+ORDER BY sequence_id ASC FOR UPDATE;
+
+-- name: GetInboxLogFiltered :many
+SELECT a.* FROM inbox_log AS a
```
neekolas, line 7: This is probably the fanciest query we have.

We need to take in an array of (`inbox_id`, `sequence_id`) pairs and get the results of each inbox that are greater than the `sequence_id`. We are currently doing this in one query per inbox, but this change makes it all happen in one query by joining against a virtual table.

pkg/mls/store/queries/queries.sql.go
```diff
@@ -0,0 +1,472 @@
+// Code generated by sqlc. DO NOT EDIT.
```
neekolas, line 1: This file is entirely generated based the the `queries.sql`

37ng (approved):
> Configuring the database with struct tags is errorprone, and there aren't great checks to make sure the struct tags actually match the schema.

How does sqlc manage db schema? From `pkg/mls/store/queries/models.go` I see that it's generating go structs based on migration history?

insipx (approved):

neekolas:
### Merge activity

* **Apr 25, 3:27 PM EDT**: @neekolas started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/xmtp/xmtp-node-go/380).

neekolas:
> How does sqlc manage db schema? From pkg/mls/store/queries/models.go I see that it's generating go structs based on migration history?

That's right. You just point it at your migrations folder and it reads all of them to assemble the schema. You can also point it at a live database if you want, but we don't have that configured since the migration folder seems to work just fine.

37ng:
> That's right. You just point it at your migrations folder and it reads all of them to assemble the schema. You can also point it at a live database if you want, but we don't have that configured since the migration folder seems to work just fine.

@neekolas So it's reversed `Bun`? `Bun` generates schema from defined go structs/tags.

PR #381, opened 2024-04-25
Migrate first store methods to sqlc

neekolas:

PR #383, opened 2024-04-25
Add InsertLog Query

insipx:


insipx:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/xmtp/xmtp-node-go/383?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#383** <a href="https://app.graphite.dev/github/pr/xmtp/xmtp-node-go/383?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#382** <a href="https://app.graphite.dev/github/pr/xmtp/xmtp-node-go/382?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#381** <a href="https://app.graphite.dev/github/pr/xmtp/xmtp-node-go/381?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#380** <a href="https://app.graphite.dev/github/pr/xmtp/xmtp-node-go/380?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

 Join @insipx and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

pkg/mls/store/queries.sql
```diff
@@ -31,11 +31,26 @@ INNER JOIN (
...
+VALUES ($1, $2, $3, $4)
+RETURNING *;
+
 -- name: InsertInboxLog :one
 INSERT INTO inbox_log (inbox_id, server_timestamp_ns, identity_update_proto)
 VALUES ($1, $2, $3)
 RETURNING sequence_id;
 
+-- name: RevokeAddressFromLog :exec
+UPDATE address_log
```
neekolas, line 45: Neat SQL here.

If this proves to be a pain, we could always just fetch the row using one query and then update it in another, filtering on the `association_sequence_id` from the first result.

But I like this more assuming it works correctly.

neekolas (approved):

pkg/mls/store/queries.sql
richardhuaaa: @neekolas and @insipx, how do we feel about enforcing that there can only be one row for each `(address, inbox_id)` tuple, via a uniqueness/primary key constraint? Then the associate and revoke operations can both do an `insert ... on conflict update` operation, updating only if the new sequence ID is higher than the existing one.

neekolas: @insipx what do you think? I don't see any blockers.

And it's been weirding me out having the `address_log` with no primary key.

PR #385, opened 2024-04-26
Continue sqlc migration

neekolas:
## tl;dr

- Migrates the rest of the queries to sqlc, including some helper queries only used in tests
- Deletes all the Bun models
- With Bun gone, I turned off the Simple Execution Mode, which allows pgx to use prepared statement caching (probably a pretty minor performance improvement, but can't hurt)

## What's next

No urgency at all on these, but at some point I'd like to get rid of the rest of Bun in the repo and remove the dependency. That would involve:

- Replacing `bun.Migrate` with another migration tool and porting over our migrations
- Replacing the `authz` database with `sqlc` queries (should be super easy, there's maybe 2 queries)


neekolas:
* **#385** <a href="https://app.graphite.dev/github/pr/xmtp/xmtp-node-go/385?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pkg/mls/store/queries.sql
```diff
@@ -131,49 +139,99 @@ INSERT INTO welcome_messages(installation_key, data, installation_key_data_hash,
...
+	id ASC;
+
+-- name: QueryGroupMessages :many
 SELECT
 	*
 FROM
 	group_messages
 WHERE
 	group_id = @group_id
+ORDER BY
```
neekolas, line 157: I'm using this fancy case statement for sorting for now. It seems to work fine, although I'm a little worried about perf. Sometimes using a case statement will prevent the query planner from using indexes. Will try and test

37ng: Curious about how you would do this. Would `pgbench` and tools that run thorough SQL transcripts be considered a practical approach?

37ng (approved):

neekolas:
@37ng pgbench is great. I was thinking just the simplest approach of running the query with and without the CASE statement through `EXPLAIN ANALYZE` (which we can do in retool) and seeing if the query plan is actually different.

37ng:
> through `EXPLAIN ANALYZE` (which we can do in retool) and seeing if the query plan is actually different.

@neekolas  That's pretty cool!

neekolas:
@37ng did the test in Retool and the query plan looks identical

PR #386, opened 2024-04-30
Add retry support for serializable transactions

neekolas:
## tl;dr

I noticed our `libxmtp` test suite was occasionally failing due to serializable transaction errors. This adds retries, which are required when using the serializable isolation level since errors are common and transient thanks to Postgres' conflict detection

### DIFF
--- a/pkg/mls/store/store.go
+++ b/pkg/mls/store/store.go
@@ -106,7 +106,7 @@ func (s *Store) PublishIdentityUpdate(ctx context.Context, req *identity.Publish
 		return nil, errors.New("IdentityUpdate is required")
 	}
 
-	if err := s.RunInTx(ctx, &sql.TxOptions{Isolation: sql.LevelSerializable}, func(ctx context.Context, txQueries *queries.Queries) error {
+	if err := s.RunInSerializableTx(ctx, 3, func(ctx context.Context, txQueries *queries.Queries) error {
 		inboxLogEntries, err := txQueries.GetAllInboxLogs(ctx, new_update.GetInboxId())
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


neekolas:
* **#386** <a href="https://app.graphite.dev/github/pr/xmtp/xmtp-node-go/386?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

pkg/mls/store/store.go
```diff
@@ -605,3 +605,20 @@ func (s *Store) RunInTx(
 	done = true
 	return tx.Commit()
 }
+
+func (s *Store) RunInSerializableTx(ctx context.Context, numRetries int, fn func(ctx context.Context, txQueries *queries.Queries) error) error {
+	var err error
+	for i := 0; i < numRetries; i++ {
+		select {
+		case <-ctx.Done():
```
insipx, line 613: noob Q: What does this select on this ctx do?

The default case returns nil when it finishes, but ctx seems like some sort of global context? When would it trigger?

insipx (approved):

37ng (approved):

pkg/mls/store/store.go
37ng: If you `ctrl+c` then `ctx.Done()` channel will receive a struct and this `select` block will abort.

neekolas: Exactly. It just stops you from continuing retries when the server is in the process of shutting down

neekolas:
The specific tests that were failing are a little worrying to me because they all used random wallets and inboxes, and executed each test serially, so they should be isolated from one another.

We may have to do some debugging around exactly what Postgres considers a "conflict" in the specific queries we are using here. Maybe it's a broader set of rows than you'd expect.

But these retries are a generally good idea regardless and we can tackle that separately.

PR #388, opened 2024-05-02
Set PermitWithoutStream to true

neekolas:
## tl;dr

- Our Rust GRPC clients have `keep_alive_while_idle` set to true, but our nodes have the `PermitWithoutStream` option defaulted to `false`. When these pings come in without an open stream, the server should be sending a GOAWAY message and disconnecting. I'm surprised it isn't causing more issues, but it's safe to enable the flag on the server so we might as well try and see if it improves mobile streaming.

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -142,8 +142,12 @@ func (s *Server) startGRPC() error {
 		grpc.Creds(insecure.NewCredentials()),
 		grpc.UnaryInterceptor(middleware.ChainUnaryServer(unary...)),
 		grpc.StreamInterceptor(middleware.ChainStreamServer(stream...)),
+		grpc.KeepaliveParams(keepalive.ServerParameters{
+			Time: 5 * time.Minute,
+		}),
 		grpc.KeepaliveEnforcementPolicy(keepalive.EnforcementPolicy{
-			MinTime: 15 * time.Second,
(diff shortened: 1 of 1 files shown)


pkg/api/server.go
```diff
@@ -142,8 +142,12 @@ func (s *Server) startGRPC() error {
 		grpc.Creds(insecure.NewCredentials()),
 		grpc.UnaryInterceptor(middleware.ChainUnaryServer(unary...)),
 		grpc.StreamInterceptor(middleware.ChainStreamServer(stream...)),
+		grpc.KeepaliveParams(keepalive.ServerParameters{
+			Time: 5 * time.Minute,
+		}),
 		grpc.KeepaliveEnforcementPolicy(keepalive.EnforcementPolicy{
-			MinTime: 15 * time.Second,
+			PermitWithoutStream: true,
+			MinTime:             15 * time.Second,
```
neekolas, line 150: Also, @nplasterer. I just added this config setting in a previous PR to turn it to 15 seconds. Before it was set to 5 minutes. I wonder how many Rust streaming problems were related to that.

alexrisch (approved):

pkg/api/server.go
nplasterer: O wow! 🤞

nplasterer (approved):

PR #389, opened 2024-05-17
Key package validation for Inbox Ids

insipx:


pkg/mls/api/v1/service.go
```diff
@@ -118,25 +118,45 @@ func (s *Service) RegisterInstallation(ctx context.Context, req *mlsv1.RegisterI
...
+		if err != nil {
+			return nil, status.Errorf(codes.InvalidArgument, "invalid identity: %s", err)
+		}
+
+		if len(results) != 1 {
+			return nil, status.Errorf(codes.Internal, "unexpected number of results: %d", len(results))
+		}
+		installationKey := results[0].InstallationKey
+		credential := results[0].Credential
+		if err = s.store.CreateInstallation(ctx, installationKey, "", []byte(credential.InboxId), req.KeyPackage.KeyPackageTlsSerialized, results[0].Expiration); err != nil {
```
insipx, line 132: I'm re-using the V3 stuff here. V3 includes some extra info like a wallet address. Coercing the InboxId into bytes and leaving wallet address blank. I'm not sure this is the behavior we want. If we create new tables I wasn't sure how that would effect everything else?

@neekolas @richardhuaaa

neekolas: We can hack in some fields as empty strings for now and then drop the fields from the DB in a migration when the old stuff is fully deprecated.

neekolas (approved):

PR #390, opened 2024-05-20
Fix issue with loop iterator

neekolas:
## tl;dr

- Fixes a [loop iterator](https://mohamed-abdel-mohaimen.medium.com/using-loop-iterator-variable-issues-in-golang-b76c17fb71b6) issue. My absolute least-favourite design quirk of Go. We should also think about upgrading to [1.22](https://go.dev/blog/loopvar-preview), which finally lets you do away with this behaviour.

## What this fixes

Previously, when querying `GetInboxIds` all results would be set to the last inbox ID.


insipx (approved):
+1 for upgrading to 1.22

37ng:
Had exact same issue, happy to see that they fixed it in 1.22

PR #394, opened 2024-05-29
Get Identity DB ready for prod

neekolas:
## tl;dr

- WIPES THE MLS/IDENTITY DATABASE
- Makes some schema changes to be more inbox_id native
- Removes the `GetIdentityUpdates` method which is now unused
- Consolidates on `mockery` for mock generation


dev/generate
```diff
@@ -3,8 +3,8 @@ set -e
 
 go generate ./...
 
-mockgen -package api github.com/xmtp/xmtp-node-go/pkg/proto/mls/api/v1 MlsApi_SubscribeGroupMessagesServer,MlsApi_SubscribeWelcomeMessagesServer > pkg/mls/api/v1/mock.gen.go
-mockgen -package mocks -source ./pkg/mlsvalidate/service.go MLSValidationService > pkg/mlsvalidate/mocks/mock.gen.go
+# Generate mocks
+mockery
```
neekolas, line 7: Steven was previously using mockgen to generate mocks. I decided to switch us back to the more standard `testify/mock` and `mockgen`. It's more widely used and more actively maintained. Otherwise very similar, other than some syntax changes.

This probably should have been a separate PR, but a lot of the things that were being mocked were also changing.

pkg/migrations/mls/20240528181851_init-schema.up.sql
```diff
@@ -0,0 +1,71 @@
+SET statement_timeout = 0;
```
neekolas, line 1: I consolidated all our MLS schema into one big migration after the drops. Just easier to read this way

pkg/migrations/mls/20240528181851_init-schema.up.sql
```diff
@@ -0,0 +1,71 @@
...
+CREATE TABLE group_messages(
+	id BIGSERIAL PRIMARY KEY,
+	created_at TIMESTAMP NOT NULL DEFAULT NOW(),
+	group_id BYTEA NOT NULL,
+	data BYTEA NOT NULL,
+	group_id_data_hash BYTEA NOT NULL
+);
+
+--bun:split
+CREATE INDEX idx_group_messages_group_id_id ON group_messages(group_id, id);
```
neekolas, line 23: This index is different from before. We were previously indexing on `(group_id, created_at)` and it never got updated when we changed the sort field on our queries.

pkg/migrations/mls/20240528181851_init-schema.up.sql
```diff
@@ -0,0 +1,71 @@
+SET statement_timeout = 0;
+
+--bun:split
+CREATE TABLE installations(
+	id BYTEA PRIMARY KEY,
+	created_at BIGINT NOT NULL,
+	updated_at BIGINT NOT NULL,
+	inbox_id BYTEA NOT NULL,
```
neekolas, line 8: We now store `inbox_id` on installations instead of `account_address`

pkg/migrations/mls/20240528181851_init-schema.up.sql
```diff
@@ -0,0 +1,71 @@
...
+	id BIGSERIAL PRIMARY KEY,
+	created_at TIMESTAMP NOT NULL DEFAULT NOW(),
+	installation_key BYTEA NOT NULL,
+	data BYTEA NOT NULL,
+	hpke_public_key BYTEA NOT NULL,
+	installation_key_data_hash BYTEA NOT NULL
+);
+
+--bun:split
+CREATE INDEX idx_welcome_messages_installation_key_id ON welcome_messages(installation_key, id);
```
neekolas, line 39: Also changed to match the sort/cursor field

pkg/migrations/mls/20240528181851_init-schema.up.sql
```diff
@@ -0,0 +1,71 @@
...
+
+--bun:split
+CREATE UNIQUE INDEX idx_welcome_messages_group_key_data_hash ON welcome_messages(installation_key_data_hash);
+
+--bun:split
+CREATE TABLE inbox_log(
+	sequence_id BIGSERIAL PRIMARY KEY,
+	inbox_id BYTEA NOT NULL,
```
neekolas, line 47: `inbox_id` is now a BYTEA field in the DB.

This will mean that any `inbox_id` values that get sent over the wire will need to be hex decodable. If you send something that isn't composed of hex characters it will error.

pkg/mls/api/v1/service.go
```diff
@@ -118,45 +118,22 @@ func (s *Service) RegisterInstallation(ctx context.Context, req *mlsv1.RegisterI
 		return nil, err
 	}
 
-	if req.IsInboxIdCredential {
```
neekolas, line 121: Removed support for legacy credentials, since we no longer even have an `account_address` field to store the resulting address.

pkg/migrations/mls/20240528181851_init-schema.up.sql
37ng: Would it make sense to change inbox_id type from String to Vec<u8> so that we don't need to handle encoding/decoding?

neekolas: Nobody else wants to deal with byte arrays, so we are handling everything as close to the db as possible.

neekolas: The frustration was that the platform SDKs were converting our byte arrays to strings immediately anyways, and then had to re-decode to pass them back in to our SDK as arguments. Having it be all strings in libxmtp just made it nice and consistent with the platform SDK. 

I actually agree with you. I like byte arrays. But the people have spoken, and we built all of libxmtp around inbox_id being a string. Would be a pain to change direction now, so we are only converting to a byte array where it matters most (shrinking the size of our DB indexes).

neekolas:
I've also now tested this with the libxmtp test suite and everything passes

pkg/migrations/mls/20240528181851_init-schema.up.sql
37ng: That makes sense, thanks!

37ng (approved):

PR #417, opened 2025-02-16
add option to marshal with protobuf

insipx:
Saves us some dependencies + request/response size for http on the rust side, since we must use `prost` anyway, this lets us disable the `json` feature in reqwest and serialize/deserialize network responses directly with protobuf using `Content-Type: application/x-protobuf` and `Accept: application/x-protobuf` http headers

### DIFF
--- a/pkg/api/server.go
+++ b/pkg/api/server.go
@@ -234,6 +234,7 @@ func (s *Server) startGRPC() error {
 func (s *Server) startHTTP() error {
 	mux := http.NewServeMux()
 	gwmux := runtime.NewServeMux(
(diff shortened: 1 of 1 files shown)


neekolas (approved):
Wow. I didn't know this option existed. Would be a big performance win to not have to actually deal with JSON in requests/responses. Have you tested this locally?

insipx:
have not tested locally yet, but it also allows us to share more code between the two implementations, esp in the new API design

it seems to be a feature that is used, though, since I originally found this on stack-overflow: https://stackoverflow.com/questions/75821690/how-can-i-get-the-grpc-gateway-rest-request-response-in-protobuf

PR #421, opened 2025-03-10
Coda/identity

codabrink:


pkg/migrations/mls/20250310145526_upgrade-identity.up.sql
```diff
@@ -0,0 +1,22 @@
...
+UPDATE address_log
+SET
+    identifier_kind = 1;
+
+ALTER TABLE address_log
+ALTER COLUMN identifier_kind
+SET
+    NOT NULL;
+
+CREATE INDEX idx_address_log_identifier_inbox_id ON address_log (identifier, identifier_kind, inbox_id);
```
codabrink, line 22: @neekolas Do you think removing and adding a new index on this table will be an issue?

pkg/migrations/mls/20240528181851_init-schema.up.sql
```diff
@@ -1,71 +1,71 @@
-SET statement_timeout = 0;
```
neekolas, line 1: Did anything actually change in this file?

pkg/migrations/mls/20250310145526_upgrade-identity.down.sql
```diff
@@ -0,0 +1,10 @@
+SET
+    statement_timeout = 0;
+
+DROP INDEX idx_address_log_identifier_inbox_id;
+
+ALTER TABLE address_log
+DROP COLUMN identifier_kind
+RENAME COLUMN identifier TO address;
+
+CREATE INDEX idx_address_log_address_inbox_id ON address_log (address, inbox_id);
```
neekolas, line 10: Can you create the index concurrently

pkg/migrations/mls/20250310145526_upgrade-identity.up.sql
neekolas: I don't think it's a huge deal because the table is relatively small. But it will block the node from coming up until it's completed.

Would be better if we could add the index concurrently, but IIRC we get errors doing that because it happens in a transaction

pkg/migrations/mls/20250310145526_upgrade-identity.up.sql
```diff
@@ -0,0 +1,22 @@
...
+
+-- Default all of the existing identifier_kinds to 1 (Ethereum)
+UPDATE address_log
+SET
+    identifier_kind = 1;
+
+ALTER TABLE address_log
+ALTER COLUMN identifier_kind
+SET
+    NOT NULL;
```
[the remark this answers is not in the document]
mkysel: I would do it like this ```
ALTER TABLE address_log ADD COLUMN identifier_kind INT DEFAULT 1;
UPDATE address_log SET identifier_kind = 1 WHERE identifier_kind IS NULL;
ALTER TABLE address_log ALTER COLUMN identifier_kind SET NOT NULL;
```

pkg/migrations/mls/20250310145526_upgrade-identity.up.sql
neekolas: From my recollection we can't use it in our migrations because the migration runner wraps everything in a transaction, and transactions can't create indexes concurrently. But my knowledge of this codebase is very out of date and maybe the issue is resolved.

You can try it. If the issue is still there every test that touches the DB will fail.

pkg/migrations/mls/20250310145526_upgrade-identity.up.sql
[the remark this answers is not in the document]
codabrink: You're right, I did write this with the assumption that the migration holds an exclusive lock. I can just be safe and use your approach.

PR #422, opened 2025-03-13
Hack to fix transaction serialization

neekolas:
## tl;dr

- Creates a new `inboxes` table in order to force duplicate transactions to retry

## What's happening here

There is a problem with the way that we acquire locks and read rows in the `PublishIdentityUpdates` flow.

Postgres transactions will retry if two concurrent transactions touch the same row. Sequences and advisory locks don't count here. When you are creating an inbox, there are no updates required. For a new inbox and wallet address, all we do is insert rows. This means that two concurrent transactions may not conflict from Postgres' perspective if they are both writing to the same inbox.

Each transaction is operating with a view of the world snapshotted from when the transaction started. So if a second transaction starts before the first has finished, both will see 0 existing updates. We need to rely on conflict detection on write to fix.

## How does this fix work
- We add a new table named `inboxes` that gets touched every time you insert an identity update. That guarantees that concurrent transactions touching the same inbox will retry.

Is this the best solution? Absolutely not. But it's one that works and we can always drop the table later if we can find a better alternative.


cameronvoell (approved):

neekolas:
### Merge activity

* **Mar 13, 3:41 PM PDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/xmtp/xmtp-node-go/422).

PR #433, opened 2025-05-01
Add gating interceptor, log whether clients are supported

richardhuaaa:
1. Adds a gating interceptor that is disabled for now
2. Log whether clients are supported or not to help us validate whether it will gate what we expect

Once we are ready, we need only flip the IS_GATING_ENABLED flag to true.

### DIFF
--- a/pkg/api/message/v1/context/context.go
+++ b/pkg/api/message/v1/context/context.go
@@ -77,9 +77,6 @@ func (ri *requesterInfo) isSupportedClient() bool {
 	if ri.LibxmtpVersion == "" || !semver.IsValid(ri.LibxmtpVersion) {
 		return true
 	}
-	if ri.LibxmtpVersion == "1.2.0-dev" {
-		return true
-	}
 	if semver.Compare(ri.LibxmtpVersion, "1.1.5") >= 0 {
 		return true
 	}
--- a/pkg/api/message/v1/context/context.go
+++ b/pkg/api/message/v1/context/context.go
@@ -2,9 +2,11 @@ package context
 
 import (
 	"context"
+	"os"
 	"strings"
 
 	"go.uber.org/zap"
+	"golang.org/x/mod/semver"
 	"google.golang.org/grpc/metadata"
 )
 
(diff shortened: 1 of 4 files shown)


pkg/api/message/v1/context/context.go
```diff
@@ -61,3 +67,22 @@ func parseVersionHeaderValue(vals []string) (name string, version string, full s
...
+func (ri *requesterInfo) isSupportedClient() bool {
+	// Only version-gate on production
+	if os.Getenv("ENV") != "production" {
+		return true
+	}
+	// Err on the side of caution for unknown versions
+	if ri.LibxmtpVersion == "" || !semver.IsValid(ri.LibxmtpVersion) {
+		return true
+	}
+	if ri.LibxmtpVersion == "1.2.0-dev" {
```
neekolas, line 80: Are this and line 83 not equivalent with the semver library we are using?

pkg/api/message/v1/context/context.go
```diff
@@ -61,3 +67,22 @@ func parseVersionHeaderValue(vals []string) (name string, version string, full s
 	}
 	return
 }
+
+func (ri *requesterInfo) isSupportedClient() bool {
+	// Only version-gate on production
+	if os.Getenv("ENV") != "production" {
```
neekolas, line 73: The perf impact shouldn't be particularly noticeable ([75ns/op](https://medium.com/@morfien101/are-environment-variables-in-go-fast-3d9dde220b6f)), but given this value doesn't change and `os.GetEnv` isn't free we should probably move this out of the function so we only have to do the lookup once on startup.

neekolas (approved):

pkg/api/message/v1/context/context.go
richardhuaaa: Ah, someone had mentioned earlier that prerelease versions came out smaller. But that's not what the spec says now that I checked. Makes sense to remove this and confirm that it's not needed via telemetry.

pkg/api/message/v1/context/context.go
richardhuaaa: I did actually check the same thing, but I found it's actually a decent amount of plumbing to get the config into everything that calls this, and I decided to keep this PR small given the sensitive timing

neekolas: Fair enough. I'm not losing sleep over 75ns

PR #450, opened 2025-06-04
Extend Welcome to include Cursor

codabrink:
### Add message cursor tracking to welcome messages in MLS protocol to extend Welcome functionality
Adds a `message_cursor` field to the `welcome_messages` table and updates the MLS API to track message position in welcome messages. The changes include:

- Database migration adds `message_cursor` column with NOT NULL constraint and default value of 0 to `welcome_messages` table in [20250604154813_add_refresh_cursor_to_welcomes.up.sql](https://github.com/xmtp/xmtp-node-go/pull/450/files#diff-779a93b8e4e26565f3e22cc02f54b389c125cfff7322c22bb583954e78d8570c)
- New stored procedure `insert_welcome_message_v3` accepts and stores the cursor value during welcome message insertion
- `SendWelcomeMessages` method in [service.go](https://github.com/xmtp/xmtp-node-go/pull/450/files#diff-6d11ce32bdc179716d251a2646ec61f8b921f7d7eecd5e50d98b9ce1592287b4) passes message cursor from client requests to database storage and includes cursor in published messages
- `MlsStore.InsertWelcomeMessage` method signature updated to accept `messageCursor` parameter of type `int64`
- Protobuf definitions updated to include `MessageCursor` field in `WelcomeMessage_V1` and `WelcomeMessageInput_V1` structs

#### 📍Where to Start
Start with the `SendWelcomeMessages` method in [service.go](https://github.com/xmtp/xmtp-node-go/pull/450/files#diff-6d11ce32bdc179716d251a2646ec61f8b921f7d7eecd5e50d98b9ce1592287b4) to see how the message cursor is extracted from requests and passed through the system.

#### Changes since #450 opened

- Marked `golang.org/x/mod` dependency as indirect in module configuration [c7248f2]
- Extended database migration rollback to include function cleanup [125cfb0]
----

_[Macroscope](https://app.macroscope.com) summarized 125cfb0._

## Summary by CodeRabbit

- **New Features**
  - Added support for a message cursor field in welcome messages, enabling tracking and retrieval using this new property.
  - API schemas for welcome messages now include the message cursor field for both input and output.

- **Bug Fixes**
  - None.

- **Tests**
  - Updated tests to cover scenarios involving the new message cursor field.

- **Chores**
  - Updated generated code and documentation to reflect the addition of the message cursor and tool version changes.


pkg/migrations/mls/20250604154813_add_refresh_cursor_to_welcomes.down.sql
```diff
@@ -0,0 +1,5 @@
+SET
+    statement_timeout = 0;
+
```
neekolas, line 3: Bun migrations need to have a `bun:split` comment between statements. Look at the other migrations for reference.

pkg/mls/store/queries/queries.sql.go
```diff
@@ -420,6 +421,7 @@ func (q *Queries) InsertWelcomeMessage(ctx context.Context, arg InsertWelcomeMes
 		arg.Data,
 		arg.InstallationKeyDataHash,
 		arg.HpkePublicKey,
+		arg.GroupRefreshStateCursor,
```
neekolas, line 424: Did you manually edit this file? I don't see the corresponding change in `queries.sql`. This file is generated.

In any case, you will need to update the postgres function for inserting a welcome message. @richardhuaaa wrote the function and can help there.

I would recommend creating a new function `insert_welcome_message_v2` rather than replacing the existing function, so that it doesn't cause errors during deployment.

pkg/migrations/mls/20250604154813_add_refresh_cursor_to_welcomes.down.sql
codabrink: Got it

dev/generate
```diff
@@ -1,12 +1,16 @@
 #!/usr/bin/env sh
 set -e
 
+BRANCH=${PROTO_BRANCH:-main}
+
+echo "Generating protos from branch ${BRANCH}..."
```
richardhuaaa, line 6: What is the goal of this change? I think it's a good thing that we currently need to be very explicit about which branch the protos are generated from, e.g. if not from main then it should be clearly demarcated

pkg/mls/api/v1/service.go
```diff
@@ -295,7 +295,8 @@ func (s *Service) SendWelcomeMessages(ctx context.Context, req *mlsv1.SendWelcom
 			insertSpan, insertCtx := tracer.StartSpanFromContext(ctx, "insert-welcome-message")
 			insertLogger := tracing.Link(insertSpan, log)
 			insertLogger.Info("inserting welcome message", zap.String("client_ip", ip), zap.Int("message_length", len(input.GetV1().Data)))
-			msg, err := s.store.InsertWelcomeMessage(insertCtx, input.GetV1().InstallationKey, input.GetV1().Data, input.GetV1().HpkePublicKey, types.WrapperAlgorithmFromProto(input.GetV1().WrapperAlgorithm))
+			msg, err := s.store.InsertWelcomeMessage(insertCtx, input.GetV1().InstallationKey, input.GetV1().Data, input.GetV1().HpkePublicKey, types.WrapperAlgorithmFromProto(input.GetV1().WrapperAlgorithm), int64(input.GetV1().MessageCursor))
```
richardhuaaa, line 298: This is not due to your change, but if `input.GetV1()` returns nil, e.g. from a malformed or malicious payload, we get a nil dereference and this method will panic. I'm not sure what happens to the server in that case.

Might be nice to add a nil-check on the line before this one, or use the proto methods for fetching properties (e.g. `input.GetV1().GetInstallationKey()` instead of `input.GetV1().InstallationKey`)

pkg/mls/store/queries/queries.sql.go
```diff
@@ -404,9 +405,9 @@ func (q *Queries) InsertInboxLog(ctx context.Context, arg InsertInboxLogParams)
 
 const insertWelcomeMessage = `-- name: InsertWelcomeMessage :one
 SELECT
-	id, created_at, installation_key, data, hpke_public_key, installation_key_data_hash, wrapper_algorithm
+	id, created_at, installation_key, data, hpke_public_key, installation_key_data_hash, wrapper_algorithm, message_cursor
 FROM
-	insert_welcome_message_v2($1, $2, $3, $4, $5)
+	insert_welcome_message_v3($1, $2, $3, $4, $5, $6)
```
richardhuaaa, line 410: This is a generated file, and I don't see a corresponding change in `queries.sql`. Was this hand edited?

That would work for this PR, but the next time somebody generates the queries, they would unknowingly break everything

dev/generate
codabrink: This is for development and defaults to main when not set.

pkg/mls/api/v1/service.go
codabrink: Done

richardhuaaa (approved):
All generated code was freshly generated after the latest changes right? If so then LGTM!

PR #496, opened 2025-07-18
Fix installation pruner and make pruners independent

mkysel:


pkg/prune/prune.go
mkysel, line 1: there is no change really, other than making sure the goroutines are independent so that each pruner can fail/report independently.

pkg/prune/prune.go
```diff
@@ -43,85 +42,96 @@ func (e *Executor) Run() error {
...
-	}
+	var (
+		wg    sync.WaitGroup
+		errCh = make(chan error, len(pruners))
+	)
 
-	var totalDeletionCount int64 = 0
+	wg.Add(len(pruners))
+
+	for _, pruner := range pruners {
```
mkysel, line 52: @coderabbitai is concurrent DB access safe here?

fbac (approved):

